PR #1, opened 2013-01-11
Added negative scan increment compatibility

felix-kolbe:
And moved duplicated loopwise incrementing line


vrabaud:
I'll pull from the other pull request you created.

PR #9, opened 2013-06-04
intensity_filter: limit intensities to avoid int overflow in cur_bucket

### LINKED ISSUE #8: Intensity Filter Breaks with Big Freaking intensities
If intensities in a laser scan are large enough, intensity filter will segfault. 

Given an intensity of 3e20 (for example), so [this line](https://github.com/ros-perception/laser_filters/blob/groovy-devel/include/laser_filters/intensity_filter.h#L96) results in an integer so large that it overflows. However the negative index results in the segfault [here](https://github.com/ros-perception/laser_filters/blob/groovy-devel/include/laser_filters/intensity_filter.h#L99)

Corrected version:

```
  int cur_bucket = (int) ((filtered_scan.intensities[i]/hist_max)*num_buckets) ;
  if (cur_bucket >= num_buckets-1 || cur_bucket < 0)
    cur_bucket = num_buckets-1 ;

  histogram[cur_bucket]++ ;
```

mintar:
closes #8

### DIFF
--- a/include/laser_filters/intensity_filter.h
+++ b/include/laser_filters/intensity_filter.h
@@ -91,6 +91,7 @@ public:
       {      
 
         filtered_scan.ranges[i] = input_scan.range_max + 1.0 ;                           // If so, then make it a value bigger than the max range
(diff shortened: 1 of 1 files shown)


vrabaud:
Sorry, it is not valid anymore, I'll comment in the bug report instead.

PR #10, opened 2013-06-04
checking if histogram entry is below 0

mateiciocarlie:
If laser scan came in with bogus intensity values (which it does in simulation) this would cause the histogram creation to segfault if trying to increment with an index below 0.

This fixes the segfault, but one could argue it could be fixed differently. Like this, bogus values show up as zeros in the histogram - maybe they should be ignored instead? The histogram seems to be used just for debug purposes anyway.

Ideally, we'd also fix sim so that it does not return bogus intensities.


vrabaud:
That code is indeed terrible ... I fixed your crash in 2da0e525b1bd2c58a5269901eb57325fece1cd9e

Please let me know if that works.

I need to check the other pull requests before releasing that.

PR #12, opened 2013-07-19
missing dependency break isolated build without --install

wjwwood:
`scan_shadows_filter.h` includes `angles/angles.h` without find_package()-ing it. Therefore it fails to find the header in isolated builds without --install.

Experienced it on hydro - did not check groovy.


wjwwood:
This has been upgraded to a pull request.

vrabaud:
thx guys

PR #14, opened 2014-02-26
Don't check the intensities

efernandez:
The intensities are not used in the range filter.
Furthermore, some laser don't have intensities ---e.g hokuyo URG-04LX-UG01---, so this fails for them.


vrabaud:
thx !

PR #16, opened 2014-03-10
Fix finding of Eigen

tfoote:
The cmake find infrastructure was moved into cmake_modules in indigo.


tfoote:
@jonbinney Will you have a few minutes to release this?

jonbinney:
Merged; will release shortly.

jonbinney:
Released: https://github.com/ros/rosdistro/pull/3438

PR #18, opened 2014-06-06
Set invalid values to NaN

atse2:


jonbinney:
This looks good - it brings laser_filters in line with http://www.ros.org/reps/rep-0117.html

Some of the tests are failing, I need to debug that before I merge.

jonbinney:
Tests passing now. Merging.

PR #19, opened 2014-06-20
remove superfluous DEPENDS

v4hn:
There never was a DEPENDS flag in add_dependencies...


vrabaud:
thx, does that create an error or something ?

v4hn:
cmake 3.0 now complains about non-existing targets in add_dependencies.
As I was just compiling ROS with this version, a number of
target-does-not-exist bugs popped up.

PR #27, opened 2015-08-10
add check inf or nan of input laser_scan intensities

aginika:
When we use `laser_filters/LaserScanIntensityFilter`with gazebo,
this will die because overflowed value is passed to array index around https://github.com/ros-perception/laser_filters/blob/indigo-devel/include/laser_filters/intensity_filter.h#L96-L99

So I added to check the value of intensities and do min compare to prevent overflowed int generating.

### DIFF
--- a/include/laser_filters/intensity_filter.h
+++ b/include/laser_filters/intensity_filter.h
@@ -93,7 +93,9 @@ public:
         filtered_scan.ranges[i] = std::numeric_limits<float>::quiet_NaN();              // If so, then make it an invalid value (NaN)
       }
 
-      int cur_bucket = (int) ((filtered_scan.intensities[i]/hist_max)*num_buckets) ;
+      if( isinf((double)filtered_scan.intensities[i]) || isnan((double)filtered_scan.intensities[i]) )
+	continue;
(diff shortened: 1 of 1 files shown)


include/laser_filters/intensity_filter.h
```diff
@@ -88,12 +88,13 @@ class LaserScanIntensityFilter : public filters::FilterBase<sensor_msgs::LaserSc
     {
       if (filtered_scan.intensities[i] <= lower_threshold_ ||                           // Is this reading below our lower threshold?
           filtered_scan.intensities[i] >= upper_threshold_)                             // Is this reading above our upper threshold?
-      {      
-
+      {
```
k-okada: Do we need this change?

aginika: No, but I think it is strange to remain this empty line, so I deleted.
I will remove this fix.

k-okada:
+1, lgtm

jonbinney:
Thanks @aginika!

furushchev:
seems need another fix: https://github.com/ros-perception/laser_filters/pull/34

PR #28, opened 2015-08-18
fix nan bug

JRombouts:
This fixes a bug in the shadow filter which can occur if there are nan values in the incoming scan.


vrabaud:
shouldn't you do something with `scan_in.ranges[j]` then ?

JRombouts:
Hmm - only adding the check on `scan_in.ranges[i]` solves my crashes. I will close this pull request for now while I figure this out :).

PR #31, opened 2015-11-18
add ScanBlobFilter

k-okada:
add filter to extract blob object in the laser scan, such as human foot or chair/table foot.

@aginika, if you have picutres for this filter please add them

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -32,6 +32,7 @@
   <run_depend>laser_geometry</run_depend>
   <run_depend>pluginlib</run_depend>
   <run_depend>angles</run_depend>
+  <run_depend>cmake_modules</run_depend>
 
   <export>
     <cpp cflags="-I${prefix}/include `rosboost-cfg --cflags`" lflags=""/>


aginika:
Like this?
39 png_convert

furushchev:
ping

include/laser_filters/scan_blob_filter.h
```diff
@@ -0,0 +1,158 @@
...
+        range_blobs.push_back(range_blob);
+    }
+    // for each blob calculate center and radius
+    for (unsigned int i = 0; i < range_blobs.size(); i++) {
+        int size = range_blobs[i].size();
+        // check center of blob
+        double center_x = 0, center_y = 0;
+        for (unsigned int j = 0; j < size; j++) {
+            double x = scan_in.ranges[range_blobs[i][j]];
+            double y = scan_in.ranges[range_blobs[i][j]] * scan_in.angle_increment;
```
jonbinney: Why is y equal to `range*angle_increment`?

k-okada: I just followed https://github.com/ros-perception/laser_filters/blob/indigo-devel/include/laser_filters/scan_shadows_filter.h#L129 line

jonbinney: scan_shadows filter is just passing that through to `getAngleWithViewpoint()`, which i guess expects that. But using it here as center_y doesn't make any sense to me.... what do you want `y` to be? If it isn't going to be the cartesian `y` coordinate, you need to document what it actually is.

k-okada: you're correct, this is not straightforward and to avoid confusion, i rewrite the code. @aginika could you check if this works ok on the real robot?

k-okada:
moved to #80

PR #32, opened 2015-11-23
scan2scan filter: only publish result if filter succeeded

v4hn:
Without this patch a tf timeout in the footprint filter
will make the module publish the input scan.


vrabaud:
thx !

v4hn:
Took me about 6 hours to debug this together with #33 .

I just rebased the other one on top of your latest merge.

Thanks for the quick response :)

PR #33, opened 2015-11-23
scan_to_scan_filter_chain: make tf filter tolerance customizable

v4hn:
0.03 is completely arbitrary and too small in my case.


v4hn:
this patch retains the API (extending it) and removes a hard-coded crucial timeout value.
It would be great if this could be merged soon, so I don't have to overlay this package locally.

jonbinney:
LGTM. Thanks @v4hn for adding this.

PR #34, opened 2015-12-07
fix: segfault on accessing out of index at histogram array

furushchev:
current codes occurs yet segmentation fault because of access to out of index on histogram array.
This PR provides:
- check if intensity value is infinity or negative, which causes out of index access.
- not compute histogram when `display_histogram` is disabled

### DIFF
--- a/include/laser_filters/intensity_filter.h
+++ b/include/laser_filters/intensity_filter.h
@@ -54,6 +54,7 @@ public:
   double lower_threshold_ ;
   double upper_threshold_ ;
   int disp_hist_ ;
+  bool disp_hist_enabled_;
 
   bool configure()
   {
@@ -64,13 +65,12 @@ public:
     getParam("upper_threshold", upper_threshold_) ;
     getParam("disp_histogram",  disp_hist_) ;
 
+    disp_hist_enabled_ = (disp_hist_ == 0) ? false : true;
+
     return true;
   }
 
-  virtual ~LaserScanIntensityFilter()
-  { 
-
-  }
+  virtual ~LaserScanIntensityFilter(){}
 
   bool update(const sensor_msgs::LaserScan& input_scan, sensor_msgs::LaserScan& filtered_scan)
   {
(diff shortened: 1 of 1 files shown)


include/laser_filters/intensity_filter.h
```diff
@@ -88,17 +88,21 @@ class LaserScanIntensityFilter : public filters::FilterBase<sensor_msgs::LaserSc
...
       }
 
-      if( isinf((double)filtered_scan.intensities[i]) || isnan((double)filtered_scan.intensities[i]) )
-	continue;
-      int cur_bucket = (int) ((std::min((double)filtered_scan.intensities[i], hist_max)/hist_max)*num_buckets) ;
-      if (cur_bucket >= num_buckets-1)
-	cur_bucket = num_buckets-1 ;
-      histogram[cur_bucket]++ ;
+
+      if (disp_hist_){
```
jonbinney: Why are you only doing this computation when `disp_hist` is non-zero? From looking at the code, I think `disp_hist` is mean to be a flag about whether or not to print out the histogram. Shouldn't the histogram be updated regardless?

include/laser_filters/intensity_filter.h
```diff
@@ -88,17 +88,21 @@ class LaserScanIntensityFilter : public filters::FilterBase<sensor_msgs::LaserSc
...
+        if( isinf((double)filtered_scan.intensities[i]) || isnan((double)filtered_scan.intensities[i]))
+          continue;
+        int cur_bucket = (int) std::max((std::min((double)filtered_scan.intensities[i], hist_max)/hist_max)*num_buckets, 0.0) ;
+        if( isinf(cur_bucket) || isnan(cur_bucket) )
```
jonbinney: `cur_bucket` is an int. How can it be inf or nan?

include/laser_filters/intensity_filter.h
furushchev: @jonbinney Thank you for reviewing. Now I clearified by using boolean on the condition.
Currently, histogram is only used for printing out, so if passed `false` to param `disp_histogram`, voting to histogram itself is not needed.

include/laser_filters/intensity_filter.h
furushchev: @jonbinney Sorry, I'm just confused. I updated the codes, and it's just simple rounding floating value to an int for checking if `cur_bucket` accesses out-of-index of `histogram` array or not.

include/laser_filters/intensity_filter.h
```diff
@@ -80,28 +80,41 @@ class LaserScanIntensityFilter : public filters::FilterBase<sensor_msgs::LaserSc
...
+        if( isinf((double)filtered_scan.intensities[i]) ||
+            isnan((double)filtered_scan.intensities[i]) )
+          continue;
+
+        // Choose bucket to vote on histogram,
+        // and check the index of bucket is in the histogram array
+        int cur_bucket = (int)(filtered_scan.intensities[i] / hist_max * num_buckets);
+        if (cur_bucket > num_buckets-1)
+          cur_bucket = num_buckets-1;
+        else if (cur_bucket < 0) cur_bucket = 0;
```
jonbinney, line 111: Is this `cur_bucket < 0` the check that actually fixes the segfault?

jonbinney:
Thanks for cleaning up the `dist_histogram` flag.

include/laser_filters/intensity_filter.h
furushchev, line 111: @jonbinney I think YES.
If `filtered_scan.intensities[i] / hist_max * num_buckets` is greater than 2147483647 (maximum of int), `cur_bucket` can be minus, which causes access to out of index of array.

jonbinney, line 111: ok, I understand now. The rest of the code looks good to me, I'll try and test and merge this tomorrow.

furushchev, line 111: @jonbinney Thank you very much for further reviewing. :+1:

jonbinney:
Code looks good and tests pass. Thanks @furushchev for contributing this!

k-okada:
great! can we release new version, so we can use fixed version as deb ?

jonbinney:
Sure, I'll try and release in the next day or two

jonbinney:
@k-okada released into indigo: https://github.com/ros/rosdistro/pull/10086

furushchev:
@jonbinney Thank you for releasing! :clap:

PR #39, opened 2016-02-29
Update CMakeLists.txt

asimay:
remove eigen dependency and tidy up cmake list reduncy

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,13 +5,12 @@ project(laser_filters)
 # Find dependencies
 ##############################################################################
 
-set(THIS_PACKAGE_ROS_DEPS cmake_modules sensor_msgs roscpp tf filters message_filters
+set(THIS_PACKAGE_ROS_DEPS sensor_msgs roscpp tf filters message_filters
   laser_geometry pluginlib angles)
 
 find_package(catkin REQUIRED COMPONENTS ${THIS_PACKAGE_ROS_DEPS})
-find_package(Boost REQUIRED COMPONENTS system signals)
-find_package(Eigen REQUIRED)
-include_directories(include ${catkin_INCLUDE_DIRS} ${EIGEN_INCLUDE_DIRS}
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -33,21 +32,21 @@ add_library(pointcloud_filters src/pointcloud_filters.cpp)
 target_link_libraries(pointcloud_filters ${catkin_LIBRARIES})
 
 add_library(laser_scan_filters src/laser_scan_filters.cpp src/median_filter.cpp src/array_filter.cpp src/box_filter.cpp)
-target_link_libraries(laser_scan_filters ${catkin_LIBRARIES} ${Boost_LIBRARIES} ${Eigen_LIBRARIES})
+target_link_libraries(laser_scan_filters ${catkin_LIBRARIES})
```
jonbinney, line 35: Why isn't the `${Boost_LIBRARIES}` needed here?

asimay, line 35: because we have catkin_depend above. DEPENDS Boost is enough.

jonbinney, line 35: The "DEPENDS Boost" doesn't appear to append the Boost library flags to `catkin_LIBRARIES` for the current package. @dirk-thomas can you confirm that this is the case?

dirk-thomas, line 35: The information passed to `catkin_package()` are used to export these to downstream packages. They don't affect the current CMake file. `catkin_LIBRARIES` only contains what has been found with `find_package(catkin ...)` above (https://github.com/ros-perception/laser_filters/pull/39/files#diff-af3b638bc2a3e6c650974192a53c7291R11).

jonbinney, line 35: @asimay can you put the `${Boost_LIBRARIES}` back into each of these lines? Thanks!

asimay, line 35: ok。sure。but i tested my changes, and test result shows my modification is right and can run correctly.
I'll change back to your style later.

Issue #29, opened 2015-08-31
Remove deprecated warning from footprint filter


stwirth:
Although there might be a "newer" version of that filter in pr2_navigation_perception, this simple filter is still useful for circular shaped robots with off-centered lasers.

jonbinney:
Agreed.  Want to submit a PR and I'll merge it?

jonbinney:
Fixed by https://github.com/ros-perception/laser_filters/pull/43

PR #43, opened 2016-03-15
Removed deprecated warning from footprint filter

naveedhd:
Ignore if I am totally wrong or inform if more changes required and I will do it.

### DIFF
--- a/include/laser_filters/footprint_filter.h
+++ b/include/laser_filters/footprint_filter.h
@@ -1,13 +1,13 @@
 /*********************************************************************
 * Software License Agreement (BSD License)
-* 
+*
 *  Copyright (c) 2008, Willow Garage, Inc.
 *  All rights reserved.
-* 
+*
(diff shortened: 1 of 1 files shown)


jonbinney:
Hi @naveedhd this was deprecated before I started maintaining laser_filters, so I don't know the whole story about why it happened. I'm assuming that it was used primarily for the PR2, and someone decided it would be better to just have a PR2 specific one (PR2LaserScanFootprintFilter)

What robot do you use it on? Do you think it is generic enough to work for many different robots?

naveedhd:
Hi @jonbinney, 

It is not used in our robots. I just followed #29 .

Actually I don't know really. My guess is it wont be used generally so should have deprecated warning.

jonbinney:
Ah yes, I forgot about that conversation entirely. I'll merge this now. Thanks @naveedhd !

PR #47, opened 2016-05-09
Replace the invalid value of scans for the footprint_filter by NaN

alain-m:
The invalid value for scans in the footprint_filter is max_range+1 while the other filters (range_filter, scan_shadows, intensity) use NaN. [REP 117](http://www.ros.org/reps/rep-0117.html) states : "This REP proposes that these values be represented by -Inf, NaN, and +Inf respectively and as defined within floating point standards."

### DIFF
--- a/include/laser_filters/footprint_filter.h
+++ b/include/laser_filters/footprint_filter.h
@@ -101,7 +101,7 @@ public:
     {
       if (inFootprint(laser_cloud.points[i])){
         int index = laser_cloud.channels[c_idx].values[i];
-        filtered_scan.ranges[index] = filtered_scan.range_max + 1.0 ; // If so, then make it a value bigger than the max range
+        filtered_scan.ranges[index] = std::numeric_limits<float>::quiet_NaN();
       }
     }
 


stwirth:
+1. Intuitively `max_range+1` would mean the laser measures something far away but it does not exactly know how far as it is more than the maximum range. That does not make sense for a _footprint_ filter.

jonbinney:
Sounds right to me. My only worry is that if someone currently has code that says `if (range[i] > max_range) { discard it } else { do something }` they may have NaN values sneaking into their program after this PR.  I'm inclined to merge it anyway but I'll wait a couple days to see if anyone has any objections. @chadrockey what do you think?

jonbinney:
Need a couple more opinions about whether this is too big a change to release into indigo.
 @haidai @kaijenhsiao

KaijenHsiao:
I would be a bit worried about existing robots expecting non-NaN values.  But if the other filters use NaN already, that makes it seem less likely to me that it would ruin someone's day.  I vote merge it.

jonbinney:
Alright, I'm going to go ahead and merge. Thanks @alain-m !

PR #51, opened 2017-04-06
Specify packages names for filters in tests

jonbinney:
This has been the right way to specify filters for a long time, and avoids accidentally using a similarly named filter from another package.


v4hn:
I'm so sorry, but I just *have to* lol this :rofl: 

I guess I know how you found it..

jonbinney:
:-)

jonbinney:
Since this is a simple change that only affects the tests and has been the "right" way to do things for a while, I'm going to go ahead and merge.

PR #53, opened 2017-07-07
Fix a small typo in one of the test cases.

allenh1:
Noticed this while patching the isnan in the prior release to work on Gentoo.


jonbinney:
Good catch! I'll test on my laptop and then merge.

PR #56, opened 2017-08-24
change range_filter to infinity for it to work with obstacle_layer

jabbenseth:
if you use the ´inf_is_valid´ parameter raytracing is still possible for
scans out of the window.
Useful for laserscanners that may deliver ranges > range_max or a more detailed discrimination on which limit is not met

### DIFF
--- a/include/laser_filters/range_filter.h
+++ b/include/laser_filters/range_filter.h
@@ -1,13 +1,13 @@
 /*********************************************************************
 * Software License Agreement (BSD License)
-* 
+*
 *  Copyright (c) 2013, JSK (The University of Tokyo).
 *  All rights reserved.
-* 
+*
 *  Redistribution and use in source and binary forms, with or without
 *  modification, are permitted provided that the following conditions
 *  are met:
-* 
+*
 *   * Redistributions of source code must retain the above copyright
 *     notice, this list of conditions and the following disclaimer.
 *   * Redistributions in binary form must reproduce the above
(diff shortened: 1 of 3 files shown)


include/laser_filters/range_filter.h
```diff
@@ -52,9 +52,26 @@ class LaserScanRangeFilter : public filters::FilterBase<sensor_msgs::LaserScan>
 
   double lower_threshold_ ;
   double upper_threshold_ ;
+  bool use_message_range_ ;
```
jonbinney: Could you rename this variable (and the ROS param name) to `use_message_range_limits`? That should make it clearer what it does.

jonbinney:
Thanks @ipa-jba ! This looks like a great added feature, and seems to be backwards compatible. I made one comment about changing a variable name; after you do that I'll merge this.

jabbenseth:
@jonbinney  there you go. 

Who keeps track of the wiki page? I can add the new interface to the documentation once the PR is merged and note that this is an "unreleased but available" interface, or do you update the wiki only with new releases?

jonbinney:
Merged - thanks @ipa-jba ! 

There's not anyone in particular who keeps track of the wiki. It would be great if you do as you suggest - add the new feature perhaps with an "added August 2017" note so that people can realize that they may need to get the package from source.

I'll create issues to remind myself to do releases soon so that everyone can get this feature from debs. Out of curiosity, which ROS distro are you primarily using?

jabbenseth:
I use indigo as well as kinetic. Thanks for taking care of the release. I will update the wiki page at work tomorrow

PR #58, opened 2017-10-25
Updated deprecated pluginlib macros to avoid warning messages

nlamprian:


jonbinney:
Thanks @nlamprian ! This PR looks good; I'm going to test it this weekend before I merge it just to be sure.

PR #59, opened 2018-01-09
Make laser_filters build for ros2 (on windows 10)

bfjelds:
Should go to a ROS2 branch, but I'm not sure how to create a PR for that :)

Working towards enabling laser_filters in ROS2. This provides ability to build and updates for ROS2 constructs (like Node vs NodeHandle, LaserScan's new namespace).

There are some gaps (TransformListener's waitForTransform, time handling, test code, parameter handling, timers), but it builds and runs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,55 +1,110 @@
-cmake_minimum_required(VERSION 2.8.3)
+cmake_minimum_required(VERSION 3.5)
 project(laser_filters)
 
 ##############################################################################
 # Find dependencies
 ##############################################################################
 
-set(THIS_PACKAGE_ROS_DEPS sensor_msgs roscpp tf filters message_filters
-  laser_geometry pluginlib angles)
+find_package(ament_cmake REQUIRED)
 
-find_package(catkin REQUIRED COMPONENTS ${THIS_PACKAGE_ROS_DEPS})
-find_package(Boost REQUIRED COMPONENTS system signals)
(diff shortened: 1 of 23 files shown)


jonbinney:
@bfjelds great work, thanks for doing this!

I can create a branch for this. @wjwwood is there a best practice for naming of ROS2 branches in ROS1 repositories?

wjwwood:
@jonbinney so far we've been just using the branch `ros2` for the equivalent of "master" for ros 2 support. Though we also have an `ardent` branch on our repositories from which we could do patch releases for the first official release. You could also use `ardent-devel` if that suited you.

My recommendation is just to use `ros2` for now, since this isn't released into ardent yet. Also the recommendation might evolve as we improve the migration guide and tools to help the ROS 1/2 stuff coexist.

jonbinney:
Thanks @wjwwood . @bfjelds I've created a ros2 branch; can you open a new PR that targets it?

wjwwood:
You can change the pr's target branch now, so no need to remake it 😉.

jonbinney:
Whoooo crazy. Do that @bfjelds  :-)

bfjelds:
Awesome!  Here's the new PR: https://github.com/ros-perception/laser_filters/pull/60.

wjwwood:
Well, it was already done 😛, but #60 is fine too.
  
screen shot 2018-01-09 at 10 57 35 am

PR #60, opened 2018-01-09
Make laser_filters build for ros2 (on windows 10) 

bfjelds:
Working towards enabling laser_filters in ROS2. This provides ability to build and updates for ROS2 constructs (like Node vs NodeHandle, LaserScan's new namespace).

There are some gaps (TransformListener's waitForTransform, time handling, test code, parameter handling, timers), but it builds and runs.


jonbinney:
Code looks good. I don't have a setup for testing this, but since certainly can't break anything (no ros2 functionality to break yet), I'm going to just merge this. Thanks @bfjelds !

tfoote:
The comment about waitForTransform being missing was flagged to me. Note that waitForTransform has been replaced by canTransform with a timeout.

PR #61, opened 2018-01-12
Remove xmlrpcpp (ROS1 conecpt) and add eigen

bfjelds:
Xmlrpcpp is not needed for the ROS2 parameter concept.

Eigen is needed for laser_geometry header.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,6 @@ project(laser_filters)
 find_package(ament_cmake REQUIRED)
 
 find_package(pluginlib REQUIRED)
-find_package(class_loader REQUIRED)
 find_package(sensor_msgs REQUIRED)
 find_package(rclcpp REQUIRED)
 find_package(tf2 REQUIRED)
@@ -22,11 +21,8 @@ find_package(rostime REQUIRED)
 set(Boost_USE_STATIC_LIBS ON)
 find_package(Boost REQUIRED COMPONENTS system signals thread filesystem)
 
-find_package(TinyXML REQUIRED HINTS "*.lib") # add hint seems to allow debug lib to be found when appropriate
-
 include_directories(include 
   ${pluginlib_INCLUDE_DIRS}
-  ${class_loader_INCLUDE_DIRS}
   ${sensor_msgs_INCLUDE_DIRS}
   ${rclcpp_INCLUDE_DIRS}
   ${tf2_INCLUDE_DIRS}
@@ -36,14 +32,11 @@ include_directories(include
   ${laser_geometry_INCLUDE_DIRS}
   ${angles_INCLUDE_DIRS}
   ${rostime_INCLUDE_DIRS}
-  ${TinyXML_INCLUDE_DIRS}
-  ${EIGEN3_INCLUDE_DIR}
   ${Boost_INCLUDE_DIRS}
 )
 
 link_directories(
   ${pluginlib_LIBRARY_DIRS}
-  ${class_loader_LIBRARY_DIRS}
   ${sensor_msgs_LIBRARY_DIRS}
   ${rclcpp_LIBRARY_DIRS}
   ${tf2_LIBRARY_DIRS}
@@ -60,7 +53,6 @@ add_definitions(-DTRANSFORM_LISTENER_NOT_IMPLEMENTED)
 
 set(${PROJECT_NAME}_LIBRARIES 
   ${pluginlib_LIBRARIES}
-  ${class_loader_LIBRARIES}
   ${sensor_msgs_LIBRARIES}
   ${rclcpp_LIBRARIES}
   ${tf2_LIBRARIES}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,8 +17,9 @@ find_package(message_filters REQUIRED)
 find_package(tf2_ros REQUIRED)
 find_package(laser_geometry REQUIRED)
 find_package(angles REQUIRED)
-find_package(xmlrpcpp REQUIRED)
 find_package(rostime REQUIRED)
+
+find_package(Eigen3 REQUIRED)
 
 set(Boost_USE_STATIC_LIBS ON)
 find_package(Boost REQUIRED COMPONENTS system signals thread filesystem)
@@ -36,9 +37,9 @@ include_directories(include
   ${tf2_ros_INCLUDE_DIRS}
   ${laser_geometry_INCLUDE_DIRS}
   ${angles_INCLUDE_DIRS}
-  ${xmlrpcpp_INCLUDE_DIRS}
   ${rostime_INCLUDE_DIRS}
   ${TinyXML_INCLUDE_DIRS}
+  ${EIGEN3_INCLUDE_DIR}
   ${Boost_INCLUDE_DIRS}
 )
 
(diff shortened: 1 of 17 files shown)


package.xml
```diff
@@ -24,7 +24,6 @@
   <build_depend>tf2_ros</build_depend>
   <build_depend>laser_geometry</build_depend>
   <build_depend>angles</build_depend>
-  <build_depend>xmlrpcpp</build_depend>
```
jonbinney, line 27: @bfjelds should eigen also be added as a build depend?

CMakeLists.txt
```diff
@@ -17,9 +17,10 @@ find_package(message_filters REQUIRED)
 find_package(tf2_ros REQUIRED)
 find_package(laser_geometry REQUIRED)
 find_package(angles REQUIRED)
-find_package(xmlrpcpp REQUIRED)
 find_package(rostime REQUIRED)
 
+find_package(Eigen3 REQUIRED)
```
jonbinney: @bfjelds is this only needed for the laser_geometry headers? If so, should that package export the necessary include directories?

bfjelds: that makes more sense!  i feel like i'm learning as i go with ament ... will try exporting from laser_geometry.

CMakeLists.txt
```diff
@@ -36,9 +35,9 @@ include_directories(include
   ${tf2_ros_INCLUDE_DIRS}
   ${laser_geometry_INCLUDE_DIRS}
   ${angles_INCLUDE_DIRS}
-  ${xmlrpcpp_INCLUDE_DIRS}
   ${rostime_INCLUDE_DIRS}
   ${TinyXML_INCLUDE_DIRS}
+  ${EIGEN3_INCLUDE_DIR}
```
jonbinney: @bfjelds actually I think that even this might not be necessary; if arment is similar to catkin, then I believe laser_geometry should export the eigen dependency such that the eigen include directories end up in laser_geometry_INCLUDE_DIRS. @dirk-thomas if laser_geometry uses eigen headers, how should it export that? Does the export_dependencies do the trick? See https://github.com/ros-perception/laser_geometry/pull/27/files#diff-af3b638bc2a3e6c650974192a53c7291R45

package.xml
bfjelds: I'm not sure :)

If I limit all references to eigen to laser_geometry, do I need a build_depend in laser_filters?

CMakeLists.txt
bfjelds: That makes sense ... makes me think that class_loader is probably not required either (shouldn't pluginlib export it as well?).

dirk-thomas: > if laser_geometry uses eigen headers, how should it export that? Does the export_dependencies do the trick?

I don't think so. If the CMake variables don't follow the "convention" - you `find_package(Eigen3)` but the variables are named `EIGEN3_` - notice the difference case - I would expect that the include directories of `laser_geometry` won't include Eigen. You might need a CMake extra file to find_package Eigen and then add its include directories to the packages include dirs.

bfjelds: Would it be better for laser_geometry to export the include dir like this:

```
ament_export_include_directories(include ${EIGEN3_INCLUDE_DIR})
```

Or for laser_filters to find_package(Eigen3)?

dirk-thomas: > Would it be better for laser_geometry to export the include dir like this:

That makes the generated CMake config file contain the absolute path. In general it is preferred to resolve the dependency "on-demand" instead.

> Or for laser_filters to find_package(Eigen3)?

Downstream packages shouldn't need to find packages which are only transitive dependencies.

bfjelds: OK.  I'm not too sure what "on-demand" means in this context, but I also tried adding something like this to laser_geometry's CMakeLists.txt:

```
# Eigen3 uses non-standard variable for
# include dirs (case and name): EIGEN3_INCLUDE_DIR.
if(NOT Eigen3_INCLUDE_DIRS)
  if (EIGEN3_INCLUDE_DIR)
    message("use EIGEN3_INCLUDE_DIR (${EIGEN3_INCLUDE_DIR}) for Eigen3_INCLUDE_DIRS")
    set(Eigen3_INCLUDE_DIRS ${EIGEN3_INCLUDE_DIR})
    message("Eigen3_INCLUDE_DIRS=${Eigen3_INCLUDE_DIRS}")
  else()
    message("Eigen3_INCLUDE_DIRS not found")
  endif()
endif()
```

In looking at how ament's export works, it seems like this "translation" should work (and anecdotally, the build works with it).

jonbinney: @bfjelds sorry I haven't had time to look deeper into this. Here's my undesrtanding so far:

- On kinetic (and other recent rosdistros, laser_geometry has used `FindPackage(eigen)` which has resolved to the ROS-specific cmake file for eigen here: https://github.com/ros/cmake_modules/blob/0.4-devel/cmake/Modules/FindEigen.cmake
- That isn't actually best practice. Since ROS Jade, packages are encouraged to directly `FindPackage(Eigen3)` as described here: http://wiki.ros.org/jade/Migration#Eigen_CMake_Module_in_cmake_modules . This is complicated because Eigen uses non standard variable names. The recommended workaround in the previous link for catkin is to explicitly add eigen's include directories to the INCLUDE_DIRS exported by the package using it.
- I think that @dirk-thomas is suggesting adding a CMake extra file to laser_geometry. That file would then be run whenever somebody does `FindPackage(laser_geometry)`, and it would call `FindPackage(Eigen3)` and add those include directories to laser_geometry_INCLUDE_DIRS.

@dirk-thomas does that summary sound right? @bfjelds I wasn't able to find any ament-specific documentation on adding a CMakeExtras file, but I think you should be able to create one for laser_geometry by following the example rcl_cpp-extras.cmake in the rcl_cpp package: https://github.com/ros2/rclcpp/tree/b81f55e5df0c3a43ba454b741e772ba77de7375b/rclcpp

Thanks for working on this @bfjelds . You're blazing a good path - working through lots of bugs since this is new.

CMakeLists.txt
```diff
@@ -17,17 +16,13 @@ find_package(message_filters REQUIRED)
 find_package(tf2_ros REQUIRED)
 find_package(laser_geometry REQUIRED)
```
bfjelds, line 17: Eigen3 include dirs. is passed through laser_geometry_INCLUDE_DIRS … that change can be found in this commit:

https://github.com/ros-perception/laser_geometry/pull/27/commits/5cd2f1c9b617c9b42fdfac9f6d54e96e823e2e9a

package.xml
jonbinney: If laser_filters doesn't directly use any eigen types or include any eigen headers, then no build depend should be needed on eigen in this package, you are right.

CMakeLists.txt
jonbinney: Awesome!

jonbinney:
I'll merge this once the corresponding PR has been merged in laser_geometry

package.xml
bfjelds: laser_filters uses headers from laser_geometry that reference eigen … so in a way, laser_filters is doing the building that requires eigen (laser_geometry is just exporting the headers).  

that gets a little "it depends on what the definition of is is" to me :)

I figure that since you can't get to laser_geometry's headers without it exposing the build_depend on eigen, the build_depend isn't needed here.  That said, I'm a little dizzy from typing that.

CMakeLists.txt
dirk-thomas: > does that summary sound right?

:+1:

PR #62, opened 2018-01-30
Reduce computation cost of ScanShadowsFilter

at-wat:
ScanShadowsFilter required a lot of CPU power mainly due to atan2.
This commit reduces computation cost of the filter.

* Remove atan2 and directly compare tangent values
* Add a test to check geometric calculation

### DIFF
--- a/include/laser_filters/scan_shadow_detector.h
+++ b/include/laser_filters/scan_shadow_detector.h
@@ -1,0 +1,79 @@
+/*********************************************************************
+* Software License Agreement (BSD License)
+* 
+*  Copyright (c) 2017, laser_filters authors
+*  All rights reserved.
+* 
+*  Redistribution and use in source and binary forms, with or without
+*  modification, are permitted provided that the following conditions
+*  are met:
+* 
+*   * Redistributions of source code must retain the above copyright
+*     notice, this list of conditions and the following disclaimer.
+*   * Redistributions in binary form must reproduce the above
+*     copyright notice, this list of conditions and the following
+*     disclaimer in the documentation and/or other materials provided
+*     with the distribution.
+*   * Neither the name of the Willow Garage nor the names of its
+*     contributors may be used to endorse or promote products derived
+*     from this software without specific prior written permission.
+* 
+*  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+*  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+*  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+*  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+*  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+*  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+*  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+*  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+*  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+*  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+*  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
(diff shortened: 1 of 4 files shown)


at-wat:
Here is a rough evaluation result of the CPU usage on processing five 2-D LIDARs' data (x6 replay of a bag) on Xeon E5-2620.

branch | CPU usage
------ | ---------
indigo-devel | 70%
optimize-shadows-filter | 20%

jonbinney:
Thanks @at-wat ! I'll find some time to review this.

include/laser_filters/scan_shadow_detector.h
```diff
@@ -0,0 +1,79 @@
...
+{
+class ScanShadowDetector
+{
+public:
+  float min_angle_tan_, max_angle_tan_;  // Filter angle threshold
+
+  void configure(const float min_angle, const float max_angle)
+  {
+    min_angle_tan_ = tanf(min_angle);
+    if (min_angle_tan_ < 0.0)
```
at-wat: This `if` is a correction for a precision loss of float around 90 degrees.
It might be better to add a comment.

include/laser_filters/scan_shadows_filter.h
```diff
@@ -82,26 +81,44 @@ class ScanShadowsFilter : public filters::FilterBase<sensor_msgs::LaserScan>
...
       return false;
     }
-    neighbors_ = 0;//default value
+    neighbors_ = 0;  // default value
     if (!filters::FilterBase<sensor_msgs::LaserScan>::getParam(std::string("neighbors"), neighbors_))
     {
       ROS_INFO("Error: ShadowsFilter was not given neighbors.\n");
     }
 
+    if (min_angle_ < 0)
```
at-wat, line 90: The PR adds a restriction of min and max angle to the shadow detection threshold.

`min_angle` larger than 90 degrees seems odd for the shadow detection; however, this might break unintended usages.
Is it better to add a parameter to switch old algorithm and the new one?

jonbinney:
Before reviewing this, I had to spend some time understanding what the filter does. Since I suck at trig, I made the following diagram:

scan_shadow_filter

That digram shows how "getAngleWithViewpoint()" computes the perpendicular angle. Here p1 and p2 are two points within the window that are being compared.

jonbinney:
@at-wat I'm still trying to fully understand the case where `min_angle` is greater than 90 degrees. From the diagram I made, I think that for two points p1 and p2 in the scan, if `getAngleWithViewpoint(r1, r2, included_angle)` is x, then `getAngleWithViewpoint(r2, r1, -included_angle)` will be `180 - x - included_angle`. So if included_angle were always positive, then a `min_angle` of greater than 90 degrees truly would make no sense, because either the perpendicular angle for p1 vs p2 would fail the test, or p2 vs p1 would fail the test.

Does that make sense? Now, actually included_angle can be negative in this case. So I don't fully understand the implications of `min_angle` being greater than 90 degrees.

jonbinney:
I wrote a quick python implementation of getanglewithviewpoint to play around with it:
```python
import numpy as np

def getAngleWithViewpoint(r1, r2, included_angle):
    return np.arctan2(r2*np.sin(included_angle), r1-r2*np.cos(included_angle))

parameters = [
        (1.0, 1.0, 0.1),
        (1.0, 1.0, -0.1)]
for r1, r2, included_angle in parameters:
    print r1, r2, included_angle, ':', getAngleWithViewpoint(r1, r2, included_angle)

```

at-wat:
Thank you for your follow up.

> So if included_angle were always positive, then a `min_angle` of greater than 90 degrees truly would make no sense, because either the perpendicular angle for p1 vs p2 would fail the test, or p2 vs p1 would fail the test.

I agree that the all points fails the test.
However, only farther points, excluding self, are deleted.
https://github.com/ros-perception/laser_filters/blob/5e76d3283dad78fc5d538986bc1789b74fb78b6a/include/laser_filters/scan_shadows_filter.h#L134
So, this filter with ~~~negative~~~ `min_angle > 90` (edited)  behaves like convex corner points detection.
I don't think this usage makes sense, anyway.
screenshot from 2018-02-09 21-30-47
(green points are filtered results with `min_angle = 91`)


> Does that make sense? Now, actually `included_angle` can be negative in this case. So I don't fully understand the implications of `min_angle` being greater than 90 degrees.

Perpendicular angle with negative `included_angle` is equivalant to `-getAngleWithViewpoint(..., fabs(included_angle))`.
In the original code, the perpendicular angle is tested with abs(), so we don't have to specially care about negative `included_angle`.
https://github.com/ros-perception/laser_filters/blob/5e76d3283dad78fc5d538986bc1789b74fb78b6a/include/laser_filters/scan_shadows_filter.h#L129
(Here, this abs() should be fabs(), BTW.)

at-wat:
@jonbinney ping
If any other comments, I'll address them.

jonbinney:
@at-wat sorry for the delay. I've been weighing the advantages of this PR with the danger of possibly changing the behavior of a system that works. Overall, I think this is an improvement. Merging.

at-wat:
Thanks!

Issue #69, opened 2019-02-28
Convention for calling filter plugins not followed in all files


YoshuaNava:
Hi,
I'm a user of the filters, and after switching to ROS Melodic I'm getting warnings about the laser filter files not specifying the parent owner of their plugins.

An example:
```
Bad filter type ScanShadowsFilter. Filter type must be of form <package_name>/<filter_name>
```

I've checked the files in your repo, and noticed that [one of the yaml files](https://github.com/ros-perception/laser_filters/blob/indigo-devel/test/test_scan_filter_chain.yaml) complies with the <package_name>/<filter_name> convention, while [most of the other files](https://github.com/ros-perception/laser_filters/tree/indigo-devel/examples) do not.

Indigo is EOL in a month, and I think this package should remain compatible with newer versions. Hence, it would be good to update these files. I'll submit a PR with the relevant changes.

PR #70, opened 2019-02-28
Fixed naming of laser filter plugins in yaml files

YoshuaNava:
This PR solves issue #69.

I just appended the prefix ```laser_filters/``` to all the filter plugins call in this package yaml files.

### DIFF
--- a/examples/footprint_filter_example.yaml
+++ b/examples/footprint_filter_example.yaml
@@ -1,5 +1,5 @@
 scan_filter_chain:
 - name: footprint_filter
-  type: LaserScanFootprintFilter
+  type: laser_filters/LaserScanFootprintFilter
   params:
     inscribed_radius: 0.325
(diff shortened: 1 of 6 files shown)


jonbinney:
@YoshuaNava thanks for this fix. I'll also forward-port it to the kinetic-devel branch.

RohanDwivedi:
Hi, @YoshuaNava  I installed the melodic branch of map_laser package from the ros melodic repo `sudo apt-get install ros-melodic-map-filter` which depends on this filter, yet the issue doesn't seem to have been resolved over there yet!  Can I understand how that can be done?
 update: fixed it, my bad, the yaml file in the package was not following the convention

https://github.com/wg-perception/people/issues/103

jonbinney:
@RohanDwivedi that's probably because this package hasn't been released on melodic since the fix was merged. So the source code in this repo has the fix, but the debian packages don't. Can you try cloning the indigo-devel version of this repo to your catkin workspace, recompiling, and see if that fixes the problem?

PR #73, opened 2019-05-15
Added error message when the filter chain failed.

peci1:


jonbinney:
Looks reasonable to me. Thanks @peci1 !

PR #74, opened 2019-05-21
scan_to_cloud_filter_chain: Make cloud channels configurable

peci1:
Allow to configure channels of the resulting point cloud.

The defaults are set so that the node is backwards-compatible.


jonbinney:
I'm triaging old issues and am going to go ahead and close this. If it is still useful it would probably need to be re-targeted to the kinetic-devel branch, since that is what is released into kinetic and later ros distros.

peci1:
Revived in #114.

PR #75, opened 2019-05-21
Added nodelet version of scan_to_cloud_filter_chain .

peci1:
The node version has been kept (hopefully without a change).


src/scan_to_cloud_filter_chain.cpp
```diff
@@ -245,18 +264,28 @@ class ScanToCloudFilterChain
...
+  ros::NodeHandle nh, pnh("~");
+  ScanToCloudFilterChain f(pnh, "");
 
   ros::spin();
 
   return (0);
 }
-
-
+#endif
```
jonbinney: Nitpick - could you add an end line to the last line of this file and nodelets.xml?

src/scan_to_cloud_filter_chain.cpp
```diff
@@ -245,18 +264,28 @@ class ScanToCloudFilterChain
...
+
+PLUGINLIB_EXPORT_CLASS(ScanToCloudFilterChainNodelet, nodelet::Nodelet)
+#else
 int
 main (int argc, char** argv)
 {
   ros::init (argc, argv, "scan_to_cloud_filter_chain");
-  ros::NodeHandle nh;
-  ScanToCloudFilterChain f;
+  ros::NodeHandle nh, pnh("~");
```
jonbinney: Technically we don't need to create the public node handle, right? Creating the private node handle should initialize all the ros stuff (i think), and the variable  `nh` isn't ever used.

src/scan_to_cloud_filter_chain.cpp
```diff
@@ -95,8 +107,9 @@ class ScanToCloudFilterChain
   bool  incident_angle_correction_;
 
   ////////////////////////////////////////////////////////////////////////////////
-  ScanToCloudFilterChain () : laser_max_range_ (DBL_MAX), private_nh("~"), filter_(tf_, "", 50),
-                   cloud_filter_chain_("sensor_msgs::PointCloud2"), scan_filter_chain_("sensor_msgs::LaserScan")
+  ScanToCloudFilterChain (ros::NodeHandle& nh, const std::string& name) :
+      laser_max_range_ (DBL_MAX), private_nh(nh), name_(name), filter_(tf_, "", 50),
```
jonbinney: Could you change the parameter name that gets passed in to the contructor from `nh` to `pnh`? Usually `nh` is the public node handle and `pnh` is the private node handle, they exist side by side, but have different values.

jonbinney (changes requested):
Couple small changes requested. After that, I'll double check that everything runs well for me and tests still pass. Thanks @peci1 !

peci1:
Thanks for having a look at the PR. I fixed the nitpicks you found.

jonbinney (approved):

jonbinney:
I have retargeted this to kinetic-devel and resolved some conflicts that created. Will merge once the ros tests pass on CI.

src/scan_to_cloud_filter_chain.cpp
```diff
@@ -260,18 +280,28 @@ class ScanToCloudFilterChain
...
 } ;
 
+#if BUILDING_NODELET
 
+class ScanToCloudFilterChainNodelet : public nodelet::Nodelet
+{
+  std::unique_ptr<ScanToCloudFilterChain> chain_;
+
+  void onInit() override {
+    chain_ = std::make_unique<ScanToCloudFilterChain>(getPrivateNodeHandle(), getName());
```
jonbinney: @peci1 looks like `std::make_unique` is only C++14 onwards. Since the kinetic-devel branch should compile on kinetic and after, could you replace that with a regular `chain_ = std::unique_ptr<ScanToCloudFilterChain>(new ...)`?

I'm planning to make a noetic branch that will take all future changes; another option would be to merge this into that branch.

peci1: I've made the requested change.

jonbinney:
Thanks @peci1 !

PR #76, opened 2019-05-21
Fix Travis and move on to Kinetic and Lunar.

peci1:


jonbinney:
@peci1 thanks, this is a much needed change! The tests should also work on melodic - could you add that?

peci1:
I'd really like to add a check for bionic/melodic, but it seems that the publicly available Travis doesn't have bionic images "yet": https://travis-ci.community/t/ubuntu-18-04-1-lts-bionic-beaver/1270/10 .

Maybe switching to ROS buildfarm would help?

jonbinney:
@peci1 argh... you are quite right. I forgot that travis tends to be a year or two behind the times. Lets stick with what we've got for now. Thanks!

Issue #78, opened 2019-06-20
Inverse Box Filter or ROI filter


Jailander:
Hi,

I was wondering why there isn't any **inverse** Box Filter?
I mean a filter that filters out all measurements outside of the box or **region of interest**, it seems that it could be done just as a copy of the Box filter plugin but changing this [line](https://github.com/ros-perception/laser_filters/blob/indigo-devel/src/box_filter.cpp#L177)
to: `if(!inBox(point)){`.
 
Is there any way of already doing this that I am missing, or is my proposed solution a contribution worth making?

jonbinney:
@Jailander i think it would be even easier to just add this feature to the existing box filter by adding an "invert_selection" parameter to the filter which negates that check. And if that param defaults to false, then adding it won't break any existing use cases. If you're interested, go ahead and make a PR.

Jailander:
Oh that is a good idea :+1: , I'll make the PR when I test it

Jailander:
Hi I made the PR as promised it works like a charm for me :+1: thank you for your suggestion

jonbinney:
Thanks @Jailander ! I'll take a look!

Jailander:
I saw the PR has been merged, thank you will close this issue :)

PR #79, opened 2019-06-25
Adding invert filter parameter to BOX filter

Jailander:
Adds a new parameter that allows to filter out the points outside the box instead of the points inside the box, as discussed in #78


jonbinney (approved):
Looks good!

PR #82, opened 2019-10-18
[fix] high_fidelity_ projection now takes laser_max_range_ into account

remco-r:
One line change, tested works.
In my own implementation (ROS Melodic Ubuntu 18) the 30m Hokuyo was somehow cutoff around 20 meters when turning on high_fidelity. This change fixes that.

### DIFF
--- a/src/scan_to_cloud_filter_chain.cpp
+++ b/src/scan_to_cloud_filter_chain.cpp
@@ -63,8 +63,7 @@ public:
   double laser_max_range_;           // Used in laser scan projection
   int window_;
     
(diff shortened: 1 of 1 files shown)


remco-r:
4a57b36  is an improvement, not a fix. The laser_max_range_ fix is most important to be pulled.

jonbinney:
@remco-r could you separate out the critical change into a separate PR? I'll merge that one and then review the rest separately.

remco-r:
@jonbinney done, https://github.com/ros-perception/laser_filters/pull/83

PR #83, opened 2019-10-22
[fix] when high fidelity true added laser_max_range_ to projection

remco-r:
One line change, tested, works.
In my own implementation (ROS Melodic Ubuntu 18) the 30m Hokuyo was somehow cutoff around 20 meters when turning on high_fidelity. This change fixes that.


jonbinney (approved):
LGTM!

PR #84, opened 2019-10-30
Nodeletise laser_filters

Patrick-Lascombe:
By porting to nodelets scan_to_scan_filter_chain and scan_to_cloud_filter_chain we save the computation time of serializing and deserializing of data.


Patrick-Lascombe:
The travis build may have failed because I added the nodelet dependencies such as ddr and nodelet. I sthere a way to solve this problem ?

jonbinney:
@Patrick-Lascombe this package is pretty low in the ros stack and i'd rather not have non-core dependencies like ddynamic reconfigure (regular dynamic reconfigure would probably be fine). Also the dynamic reconfigure changes shouldn't be in the same PR as the nodelet changes.

What were the performance problems you were seeing that you're solving with this PR? What CPU usage levels for what number of laser scans per second? And how much did that change after these changes to the code?

Patrick-Lascombe:
Sorry for letting this PR aside for this long.
Since the default branch has been moved to kinetic-devel, I've sync my fork with the upstream and moved my modification to the kinetic-devel branch.
Hence I'll close this PR and open another one on the correct branch, with the modifications asked and the CPU usage levels

PR #85, opened 2019-11-12
split ROS filter chain from node and add nodelets

c-andy-martin:
Split the ROS filter chain functionality from the node into its own
class for both the scan-to-scan and scan-to-cloud nodes.

Add nodelets for both using the new filter chain classes.


jonbinney:
I'm triaging old issues and am going to go ahead and close this. If it is still useful it would probably need to be re-targeted to the kinetic-devel branch, since that is what is released into kinetic and later ros distros.

PR #88, opened 2020-03-11
feat: PolygonFilter

reinzor:
Filters LaserScan points inside or outside a specified polygon. This
filter also exposes a dynamic reconfigure server for reconfiguration.


jonbinney:
This sounds very useful. I'll try to review the code this weekend.

reinzor:
hi @jonbinney , did you have time to look into this?

jonbinney:
@reinzor thanks for the reminder - let me look at it today.

jonbinney:
@reinzor could you retarget this PR to the kinetic-devel branch? I'm now releasing kinetic and melodic from the kinetic-devel branch, and only indigo from the indigo-devel branch.

That will also make it OK to use C++11, since starting in kinetic that is supported: https://www.ros.org/reps/rep-0003.html#kinetic-kame-may-2016-may-2021

reinzor:
sure

PR #89, opened 2020-03-13
feat(scan_shadows_filter): Parameter to remove shadow start point

reinzor:
Required for laser scanners that also show the veiling effect in front
of physical objects.

### DIFF
--- a/include/laser_filters/scan_shadows_filter.h
+++ b/include/laser_filters/scan_shadows_filter.h
@@ -55,6 +55,7 @@ public:
   double laser_max_range_;        // Used in laser scan projection
   double min_angle_, max_angle_;  // Filter angle threshold
   int window_, neighbors_;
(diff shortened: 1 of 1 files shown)


jonbinney:
@reinzor thanks for doing this! What model of laser scanner are you seeing this with? Do you have a screenshot of the laser data so I understand what this looks like?

reinzor:
>  What model of laser scanner are you seeing this with?

Sick Nanoscan3

> Do you have a screenshot of the laser data so I understand what this looks like?

I don't have access to a real robot but it looks as follows:
image

jonbinney (approved):
LGTM. Default behavior is same as before, so this shouldn't break anyone's robot.

jonbinney:
Ah yes, I've seen something similar with the microscan3. I'll try this out!

jonbinney:
@wongkaiweng FYI

reinzor:
FYI, a picture of the real scanner data:

image

PR #90, opened 2020-03-18
feat(ScanShadowsFilter): add dynamic reconfigure

YannickdeHoop:


jonbinney:
Could you retarget this to kinetic-devel branch? I'm going to release melodic and kinetic from that branch from now on. I'm separating it from the indigo-devel branch because iindigo does not support c++11.

reinzor:
sure

PR #91, opened 2020-03-25
Feat: polygon filter

reinzor:
Filters LaserScan points inside or outside a specified polygon. This
filter also exposes a dynamic reconfigure server for reconfiguration.

Follow up of https://github.com/ros-perception/laser_filters/pull/88


CMakeLists.txt
```diff
@@ -1,12 +1,14 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(laser_filters)
 
+add_compile_options(-std=c++11)
```
jonbinney: Could you change this to `set(CMAKE_CXX_STANDARD 11)`?

CMakeLists.txt
```diff
@@ -1,12 +1,14 @@
...
+
 ##############################################################################
 # Find dependencies
 ##############################################################################
 
 set(THIS_PACKAGE_ROS_DEPS sensor_msgs roscpp tf filters message_filters
-  laser_geometry pluginlib angles)
+  laser_geometry pluginlib angles dynamic_reconfigure)
```
jonbinney: Thanks for using dynamic reconfigure - it will make this much easier to use!

src/polygon_filter.cpp
```diff
@@ -0,0 +1,412 @@
...
+      c = !c;
+  }
+  return c;
+}
+
+
+void LaserScanPolygonFilter::reconfigureCB(laser_filters::PolygonFilterConfig& config, uint32_t level)
+{
+  invert_filter_ = config.invert;
+  polygon_ = makePolygonFromString(config.polygon);
```
jonbinney: It looks like `makePolygonFromString()` will return an empty polygon if the string isn't properly formatted. Could that cause problems? If `invert` is set to true, would an empty polygon filter out all laser points?

test/test_polygon_filter
```diff
@@ -0,0 +1,45 @@
...
+import math
+import unittest
+
+import rospy
+import rostest
+from sensor_msgs.msg import LaserScan
+from std_msgs.msg import Header
+
+
+class TestPolygonFilter(unittest.TestCase):
```
jonbinney, line 17: Thanks for adding a test!

jonbinney (commented):
I left a couple small comments and questions. I still need to do a second pass, look more deeply at the logic in polygon_filter.cpp, and try this out on my machine.

src/polygon_filter.cpp
reinzor: Yes, this is intended. What behavior do you suggest otherwise?

jonbinney: I haven't thought this through deeply, but what do you think of:
- on startup, if there is no valid polygon string, die
- if someone accidentally inputs an invalid polygon using dynamic reconfigure, keep the old polygon.

Mostly I'm trying to avoid the case where a obstacle avoidance starts ignoring all laser points.

src/polygon_filter.cpp
```diff
@@ -0,0 +1,413 @@
...
+        point.x = vvf[ i ][ 0 ];
+        point.y = vvf[ i ][ 1 ];
+        point.z = 0;
+        polygon.points.push_back(point);
+      }
+      else
+      {
+        ROS_ERROR("Points in the polygon specification must be pairs of numbers. Found a point with %d numbers.",
+                   int(vvf[ i ].size()));
+        return last_polygon;
```
jonbinney, line 227: :+1:

jonbinney (approved):

PR #92, opened 2020-03-25
feat(ScanShadowsFilter): add dynamic reconfigure

reinzor:
Follow up of https://github.com/ros-perception/laser_filters/pull/90


include/laser_filters/scan_shadows_filter.h
```diff
@@ -67,6 +74,12 @@ class ScanShadowsFilter : public filters::FilterBase<sensor_msgs::LaserScan>
   /**@b Configure the filter from XML */
   bool configure()
   {
+    ros::NodeHandle private_nh("~" + getName());
+    dyn_server_.reset(new dynamic_reconfigure::Server<laser_filters::ScanShadowsFilterConfig>(own_mutex_, private_nh));
+    dynamic_reconfigure::Server<laser_filters::ScanShadowsFilterConfig>::CallbackType f;
+    f = boost::bind(&laser_filters::ScanShadowsFilter::reconfigureCB, this, _1, _2);
+    dyn_server_->setCallback(f);
+
```
jonbinney, line 82: Are the `getParam()` calls below still needed? Dynamic reconfigure should read its initial values from the params anyway, right? Or did the variable names change?

reinzor: > Are the getParam() calls below still needed? Dynamic reconfigure should read its initial values from the params anyway, right? Or did the variable names change?

If we want backwards compatibility yes since all `filter` instances read their parameters from the `param` section of the filter defined in the chain. Dynamic reconfigure also sets is parameters in a separate namespace (cannot be done in another way imo) but initializes with params defined the chain.

IMO, this is a limitation of the filter chain but I think having these parameters dynamically reconfigurable is a must have features for these kind of filters.

jonbinney: Yes, that makes sense - thanks for keeping it backwards compatible!

CMakeLists.txt
```diff
@@ -1,12 +1,14 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(laser_filters)
 
+add_compile_options(-std=c++11)
```
jonbinney: Could you change this to `set(CMAKE_CXX_STANDARD 11)`

cfg/ScanShadowsFilter.cfg
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python
+#
+# Copyright (c) 2020, Eurotec, Netherlands
+# All rights reserved.
+#
+# \author Yannick de Hoop
```
jonbinney, line 35: Could you put a specific BSD-style license header in this file?

include/laser_filters/scan_shadows_filter.h
```diff
@@ -111,13 +124,33 @@ class ScanShadowsFilter : public filters::FilterBase<sensor_msgs::LaserScan>
...
+    param_config.window = window_;
+    param_config.neighbors = neighbors_;
+    param_config.remove_shadow_start_point = remove_shadow_start_point_;
+    dyn_server_->updateConfig(param_config);
+
     return true;
   }
 
+  void reconfigureCB(ScanShadowsFilterConfig& config, uint32_t level)
+  {
```
jonbinney, line 143: The reconfigure callback can be called from a different thread than the main processing code. Could you add a lock so that the parameters are never used in a half-updated state?

jonbinney (changes requested):
Overall looks very good - I've made a couple small change requests. Once those are resolved I'll approve and merge.

include/laser_filters/scan_shadows_filter.h
jonbinney, line 143: You should also grab `own_mutex_` here, right?

reinzor: This is being handled by the reconfigure server as far as I can see: https://github.com/ros/dynamic_reconfigure/blob/melodic-devel/include/dynamic_reconfigure/server.h#L216

jonbinney: Good point! I hadn't realized it used a mutex that was passed in to it.

jonbinney (approved):

PR #93, opened 2020-03-27
New filter for radius outlier removal

nlimpert:
This PR adds a new filter inspired by PCL's [`RadiusOutlierRemoval`](http://pointclouds.org/documentation/tutorials/remove_outliers.php) to allow a radius-based filtering based on plain distances between points.
I found this useful to overcome sensor noise being problematic specially for navigation purposes, where false-positive measurements are marked as obstacles. In contrast I couldn't figure out how to properly configure the `ScanShadowFilter` to achieve a robust filtering of generic sensor noise.

The implementation is based on (and also inspired by) the `ScanShadowFilter`.

The `radius_search` parameter specifies the maximum allowed distance of a neighboring point to be considered as a valid neighbor.
The parameter `min_neighbors` specifies how many neighbors have to be found to consider the point to remain in the output.
The `window` parameter specifies how many consecutive measurements to take into account.

The following screenshot shows a sample of a robot running the new filter:
Screenshot of robot running radius_outlier_filter
The laser measurements colored in magenta represent the raw, unfiltered laser while the cyan ones are the filtered measurements.
The configuration of this scenario is the same as found in the example - being:
```
    radius_search: 0.4
    min_neighbors: 10
    window: 10
```

I have also created a video demonstrating the new filter:
https://fh-aachen.sciebo.de/s/Mf9QgE8SMC8y3Aw
Note that in this video RViz is configured with `Decay Time = 2` for the filtered laser to highlight the result.

### DIFF
--- a/cfg/RadiusOutlierFilter.cfg
+++ b/cfg/RadiusOutlierFilter.cfg
@@ -42,6 +42,6 @@ gen = ParameterGenerator()
 
 gen.add("radius_search", double_t, 0, "Maximum allowed point distance", 0.4, 0, 10.0)
 gen.add("window", int_t, 0, "Number of consecutive measurements to consider", 1, 1, 100)
-gen.add("min_neighbors", int_t, 0, "Minimum number of neighbors", 20, 1, 100)
+gen.add("min_neighbors", int_t, 0, "Minimum number of neighbors", 20, 0, 100)
 
 exit(gen.generate(PACKAGE, "laser_filters", "RadiusOutlierFilter"))
--- a/cfg/RadiusOutlierFilter.cfg
+++ b/cfg/RadiusOutlierFilter.cfg
@@ -1,0 +1,47 @@
+#!/usr/bin/env python
+#
+# Software License Agreement (BSD License)
+#
+# Copyright (c) 2020, Nicolas Limpert
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above
+#    copyright notice, this list of conditions and the following
+#    disclaimer in the documentation and/or other materials provided
+#    with the distribution.
+#  * Neither the name of the TNO IVS nor the names of its
+#    contributors may be used to endorse or promote products derived
+#    from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+# FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+# COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+# INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+#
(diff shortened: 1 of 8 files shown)


nlimpert:
@jonbinney would you have the time for a review? :)

jonbinney:
Will try to get to this tomorrow!

cfg/RadiusOutlierFilter.cfg
```diff
@@ -0,0 +1,47 @@
...
+
+from dynamic_reconfigure.parameter_generator_catkin import *
+
+PACKAGE = "laser_filters"
+
+gen = ParameterGenerator()
+
+gen.add("radius_search", double_t, 0, "Maximum allowed point distance", 0.4, 0, 10.0)
+gen.add("window", int_t, 0, "Number of consecutive measurements to consider", 1, 1, 100)
+gen.add("min_neighbors", int_t, 0, "Minimum number of neighbors", 20, 1, 100)
```
jonbinney: "neighbors" doesn't include the point being tested, right? Would it make sense to make the min value for min_neighbors zero instead of 1? That would effectively turn off the filter, and would be useful during testing of the filter.

examples/radius_outlier_example.launch
```diff
@@ -0,0 +1,6 @@
+<launch>
+<node pkg="laser_filters" type="scan_to_scan_filter_chain" output="screen" name="laser_filter">
+      <!--remap from="scan" to="base_scan" /-->
```
jonbinney: Could you remove the commented out remap?

include/laser_filters/radius_outlier_detector.h
```diff
@@ -0,0 +1,101 @@
...
+    // Apply cos² + sin² = 1:
+    // d² = r2² + r1² - 2 * r1 * r2 * (cos(phi1) * cos(phi2) + sin(phi1) * sin(phi2))
+    //
+    // Note the following:
+    // cos(phi1) * cos(phi2) = 1/2 * (cos(phi1 - phi2) + cos(phi1 + phi2))
+    // sin(phi1) * sin(phi2) = 1/2 * (cos(phi1 - phi2) - cos(phi1 + phi2))
+    //
+    // cos(phi1) * cos(phi2) + sin(phi1) * sin(phi2) = cos(phi1 - phi2)
+    //
+    // Finally, phi1 - phi2 is our included_angle.
```
jonbinney, line 83: Nice explanation!

include/laser_filters/radius_outlier_filter.h
```diff
@@ -0,0 +1,190 @@
...
+  void reconfigureCB(RadiusOutlierFilterConfig& config, uint32_t level)
+  {
+    radius_search_ = config.radius_search;
+    window_ = config.window;
+    min_neighbors_ = config.min_neighbors;
+
+    radius_detector_.configure(radius_search_);
+  }
+
+  ////////////////////////////////////////////////////////////////////////////////
```
jonbinney: Could you remove the `/////////` lines from this file? I don't think they help with readability.

include/laser_filters/radius_outlier_filter.h
```diff
@@ -0,0 +1,190 @@
...
+   */
+  bool update(const sensor_msgs::LaserScan& scan_in, sensor_msgs::LaserScan& scan_out)
+  {
+    boost::recursive_mutex::scoped_lock lock(own_mutex_);
+    // copy across all data first
+    scan_out = scan_in;
+
+    std::set<int> indices_to_delete;
+    // For each point in the current line scan
+    for (unsigned int i = 0; i < scan_in.ranges.size(); i++)
```
jonbinney: In cases like this where an index variable is going to be used in math expressions, I prefer to just make it an int, and change the upper limit to `i < static_cast<int>(scan_in.ranges.size())`. Not a required change though, since it looks like you cast it correctly down below.

jonbinney (changes requested):
Requesting a couple of minor tweaks, then I'll merge. Thanks @nlimpert !

cfg/RadiusOutlierFilter.cfg
nlimpert: Yea good point - thanks for the info.
Addressed in 606ddb79f1d77afe5fbdbff3573363ac8896179e.

examples/radius_outlier_example.launch
nlimpert: Done in b88b9751fb7d5fa57975c694c050c17ea541f938.

include/laser_filters/radius_outlier_detector.h
nlimpert: Thanks! :slightly_smiling_face:

include/laser_filters/radius_outlier_filter.h
nlimpert: Done in 89a46899b9dbf5643e36374263469c11b336f263.

include/laser_filters/radius_outlier_filter.h
nlimpert: Thanks for the note - you're right. I think it's also better to change to C++-style casts in general.
I've done so in c06b709b14027d1bde7be75ed7dc2e16a8f2ece5.

jonbinney (approved):

nlimpert:
Thanks for the review!
Is there a chance for a new release including this? :)

jonbinney:
Sure - are you on melodic?

nlimpert:
> Sure - are you on melodic?

Yes. Thanks a lot!

jonbinney:
Release 1.8.10 created: https://github.com/ros/rosdistro/pull/24413

PR #96, opened 2020-04-08
feat(IntensityFilter): Dynamic reconfigure and optionally override intensity

reinzor:
- Dynamic reconfigure for IntensityFilter
- Removed printing of histogram (we can use dynamic reconfigure for tuning now)
- Moved code to cpp file instead of header
- Option to configure whether to override the range to NaN
- Option to configure whether to override the intensity values to 0 (not intense) and 1 (intense)

_This refactor is backwards compatible so it would not break things for running configurations. Except for the API/ABI stability._

### DIFF
--- a/include/laser_filters/intensity_filter.h
+++ b/include/laser_filters/intensity_filter.h
@@ -2,6 +2,7 @@
 * Software License Agreement (BSD License)
 * 
 *  Copyright (c) 2008, Willow Garage, Inc.
+*  Copyright (c) 2020, Eurotec B.V.
 *  All rights reserved.
 * 
 *  Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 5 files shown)


reinzor:
Any time to review this PR? Thanks :)

src/intensity_filter.cpp
```diff
@@ -0,0 +1,112 @@
...
+  dynamic_reconfigure::Server<IntensityFilterConfig>::CallbackType f;
+  f = boost::bind(&LaserScanIntensityFilter::reconfigureCB, this, _1, _2);
+  dyn_server_->setCallback(f);
+
+  getParam("lower_threshold", config_.lower_threshold);
+  getParam("upper_threshold", config_.upper_threshold);
+  getParam("invert", config_.invert);
+
+  getParam("filter_override_range", config_.filter_override_range);
+  getParam("filter_override_intensity", config_.filter_override_intensity);
```
jonbinney, line 61: Doesn't dynamic_reconfigure read the ROS parameters and call your callback for you? Do you really need to manually read all the variables and call updateConfig here?

include/laser_filters/intensity_filter.h
```diff
@@ -30,103 +31,32 @@
...
-    // Display Histogram
-    if (disp_hist_enabled_)
-    {
-      printf("********** SCAN **********\n") ;
-      for (int i=0; i < num_buckets; i++)
-      {
-        printf("%u - %u: %u\n", (unsigned int) hist_max/num_buckets*i,
-                                (unsigned int) hist_max/num_buckets*(i+1),
-                                histogram[i]) ;
-      }
```
jonbinney, line 125: +1 for removing all this printf'd histogram stuff!

src/intensity_filter.cpp
reinzor: That is correct but since we would like to be backwards compatible with the filter chain parameter implementation, we have to add these `getParam` calls in order to ensure that old configs still work. Parameters are read from the `params` namespace in the defined filter array item.

jonbinney: Got it, thanks!

jonbinney (approved):

PR #102, opened 2021-01-28
Added support for laserscanners that spin clockwise 

teundeplanque:
Added support for laserscanners that spin clockwise to the angular bounds filter (i.e. angle increment is negative)

### DIFF
--- a/include/laser_filters/angular_bounds_filter.h
+++ b/include/laser_filters/angular_bounds_filter.h
@@ -73,28 +73,53 @@ namespace laser_filters
         unsigned int count = 0;
         //loop through the scan and truncate the beginning and the end of the scan as necessary
         for(unsigned int i = 0; i < input_scan.ranges.size(); ++i){
-          //wait until we get to our desired starting angle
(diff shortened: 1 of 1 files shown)


include/laser_filters/angular_bounds_filter.h
```diff
@@ -73,28 +73,53 @@ namespace laser_filters
...
-            if(current_angle + input_scan.angle_increment > upper_angle_){
-              break;
+              //check if we need to break out of the loop, basically if the next increment will put us over the threshold
+              if(current_angle + input_scan.angle_increment > upper_angle_){
+                break;
+              }
+              current_angle += input_scan.angle_increment;
+            }
+          }
+          else{ //the laserscanner turns clockwise
```
jonbinney, line 99: This is a fair amount of repeated code, but in this case I think that is OK - it avoids changing the code for the positive case, so it is unlikely to break anything.

jonbinney (approved):
Code looks good. Seems unlikely to break existing cases, and adds important functionality.

PR #104, opened 2021-02-01
Porting speckle filter to ROS2

Patrick-Lascombe:
Porting speckle filter to ROS2 which was missing for now

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@ project(laser_filters)
 ##############################################################################
 
 find_package(ament_cmake_auto REQUIRED)
-find_package(pcl_conversions REQUIRED)
+
 ament_auto_find_build_dependencies()
 
 ##############################################################################
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,7 @@ project(laser_filters)
 ##############################################################################
 
 find_package(ament_cmake_auto REQUIRED)
+find_package(pcl_conversions REQUIRED)
 ament_auto_find_build_dependencies()
 
 ##############################################################################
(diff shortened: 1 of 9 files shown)


jonbinney:
Looks like the CI is only failing because it tries to test the code with ROS1. I should update the CI jobs at some point....

@Patrick-Lascombe it looks like you copied the files over without preserving the commits. Could you instead cherry-pick the commits over so that we can tell where this came from in the future?

Patrick-Lascombe:
Yes I copied the file from the ROS1 branch and "translated" it into ROS2, I don't know if there is a way to cherry pick the ROS 1 speckle filter file to make it appear in the commits ?

Patrick-Lascombe:
Sorry I left this on the side for a while now, here is the [permalink](https://github.com/ros-perception/laser_filters/blob/2490b8aff7ac9be93e1ab68581146ca0c65eaa5a/src/speckle_filter.cpp) of the file I copied and translated to ROS2. 
I hope it is enough for the tracking of the speckle filter origin.

jonbinney:
@Patrick-Lascombe yes you can copy over the original commit(s).  Please use the `git cherry-pick` command to copy over the original speckle filter commit(s), then reapply your change commits.

pmusau17:
@Patrick-Lascombe wondering if you're around to do the cherry pick? I tried to do it on my end but think that I might be missing the final port of this [file](https://github.com/ros-perception/laser_filters/blob/2490b8aff7ac9be93e1ab68581146ca0c65eaa5a/src/speckle_filter.cpp)

Patrick-Lascombe:
Sorry, I've had a lot of work lately and I couldn't take the time. I think that this file is from where I began the porting.
If it is not too urgent I'll try to cherry-pick and port again to ROS2 by the end of next week

jonbinney:
I went ahead and made a new PR that cherry-picks the original speckle filter commits and the update commits from this PR: https://github.com/ros-perception/laser_filters/pull/122

I'll do some testing tomorrow before I merge it.

PR #105, opened 2021-02-02
Add circle sector sharp filter

isjfk:
Filter points inside (or outside) of a circle sector sharp.
In my recent ROS setup I need to remove points inside a circle sector area. At first I think it can be archieved by combine LaserScanAngularBoundsFilterInPlace & LaserScanRangeFilter to get the overlapped area, however it turns out to be very difficult if not possible. So I write a new filter to do the work.
It also support to set angle_min > angle_max to work like a combination of LaserScanAngularBoundsFilter & LaserScanRangeFilter.
A invert parameter is also added to invert the filter area.


jonbinney:
@isjfk could you explain a bit more why the same thing can't be done with just the angular bounds filter and range filter?

isjfk:
> @isjfk could you explain a bit more why the same thing can't be done with just the angular bounds filter and range filter?

Hi @jonbinney ,

It's a little long story. My robot have a Slamtec RPLIDAR-A1 laser scanner. It's a $90 cost saving choice, probably because of that it constantly detect some random points which should not exists. I never test a better one so I don't know how it work out in expensive scanners. The env have a lot of glass walls and reflective objects, which may be another reason. The consequence is the robot frequently stop in sudden, then rotate try to avoid obstacle for no reason, even in a clean space (because the scanner "see" some not existing things in close front).
We have another Realsense D435i depth camera which works very well, no phantom points. However it have a very limited angle (-0.65, 0.65) and range (<3). The idea is if we remove points from laser scanner in angle (-0.65, 0.65) && range (<3), then combine with scan data from depth camera, it should bypass the phantom points issue. The result is actually quite good, robot walks very smooth now, the sudden stop for no reason issue almost gone.

I tried with existing filters but can't find a solution for my requirements. If I use a LaserScanAngularBoundsFilterInPlace:
```
- name: angle
  type: laser_filters/LaserScanAngularBoundsFilterInPlace
  params:
    lower_angle: -0.65
    upper_angle: 0.65
```
It removes everything in angle (-0.65, 0.65), however I want to keep points in range (>3) even they are in angle (-0.65, 0.65). The LaserScanRangeFilter have similar issue, it never take angle into consideration.

The most close configure I can figure out is:
```
- name: angle
  type: laser_filters/LaserScanAngularBoundsFilter
  params:
    lower_angle: -0.65
    upper_angle: 0.65
- name: range
  type: laser_filters/LaserScanRangeFilter
  params:
    lower_threshold: 0
    upper_threshold: 3
```
It will keep all points in angle (-0.65, 0.65) && range (<3), remove all other points. The sharp is exact the area I want, but I want to do the invert action -- I want to remove points in this area, not keeping them. I know there are undocumented "invert" param in some filters (though not in angle filters and range filters), but I couldn't find a way to invert a filter chain.

That's the reason I decide to write a new filter. Please correct me if there are better way to get the same result.

Best Regards,
Jimmy

jonbinney:
Thanks for the explanation, that makes perfect sense! I guess the general problem is that a chain of filters effectively means "accept points which match this filter AND this filter AND this other filter". We don't have any way to say "accept points which match this filter OR this other filter". Sorry I've been slow to respond; I'll try to review this in the net few days.

isjfk:
Thanks @jonbinney . I also agree if there are AND, OR, NOT and group logics in the filter chain, it will be much more flexible. It will make very sophisticated result be possible from composite some very basic filters.

PR #107, opened 2021-03-03
Dev angular interval filter

ulassbin:
Implemented laser angular interval filter and added yaml description in test/test_scan_filter_chain.yaml.

Angular interval filter takes a parameter interval sets in the form of:
    interval_sets: "[[0,0.5],[2,3.2],[5,6.5]]"

which will crop the laser scan between the ranges 
0 to 0.5 
2 to 3.2
5 to 6.5

Also this parameter can be changed from dynamic parameter server. 
It is disabled by default. Can be enabled via dynamic parameter server.

### DIFF
--- a/cfg/AngularIntervalFilter.cfg
+++ b/cfg/AngularIntervalFilter.cfg
@@ -1,0 +1,46 @@
+#!/usr/bin/env python
+#
+# Software License Agreement (BSD License)
+#
+# Copyright (c) 2020, Eurotec, Netherlands
+# All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+#  * Redistributions of source code must retain the above copyright
+#    notice, this list of conditions and the following disclaimer.
+#  * Redistributions in binary form must reproduce the above
+#    copyright notice, this list of conditions and the following
+#    disclaimer in the documentation and/or other materials provided
+#    with the distribution.
+#  * Neither the name of the TNO IVS nor the names of its
+#    contributors may be used to endorse or promote products derived
+#    from this software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+# "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+# LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+# FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+# COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+# INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+#
+# \author Yannick de Hoop
+
+from dynamic_reconfigure.parameter_generator_catkin import *
+
+PACKAGE = "laser_filters"
+
+gen = ParameterGenerator()
+
+gen.add("interval_sets",    str_t,    0, "A string parameter",  "[[-1,1],[-1.2,-1.1],[0.5,0.8]]")
+gen.add("enable", bool_t, 0, "Whether the filter is enabled or not", False)
+
+exit(gen.generate(PACKAGE, "laser_filters", "AngularIntervalFilter"))
--- a/include/laser_filters/angular_interval_filter.h
+++ b/include/laser_filters/angular_interval_filter.h
@@ -1,0 +1,223 @@
+/*********************************************************************
+* Software License Agreement (BSD License)
+*
+*  Copyright (c) 2013, JSK (The University of Tokyo).
+*  All rights reserved.
+*
+*  Redistribution and use in source and binary forms, with or without
+*  modification, are permitted provided that the following conditions
+*  are met:
+*
+*   * Redistributions of source code must retain the above copyright
+*     notice, this list of conditions and the following disclaimer.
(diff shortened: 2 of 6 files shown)


cfg/AngularIntervalFilter.cfg
```diff
@@ -0,0 +1,46 @@
...
+#
+# \author Yannick de Hoop
+
+from dynamic_reconfigure.parameter_generator_catkin import *
+
+PACKAGE = "laser_filters"
+
+gen = ParameterGenerator()
+
+gen.add("interval_sets",    str_t,    0, "A string parameter",  "[[-1,1],[-1.2,-1.1],[0.5,0.8]]")
```
jonbinney, line 43: `interval_sets` could use a better help string. Also what do you think about shortening the name to just `intervals`? `sets` sounds discrete.

cfg/AngularIntervalFilter.cfg
```diff
@@ -0,0 +1,46 @@
...
+gen.add("enable", bool_t, 0, "Whether the filter is enabled or not", False)
```
jonbinney, line 44: Maybe remove enable? I can see how it saves a bit of yaml editing during testing, but it might confuse users since none of the other filter types have this option.

cfg/AngularIntervalFilter.cfg
```diff
@@ -0,0 +1,46 @@
...
+# BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+# LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+# CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+# LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+# ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+#
+# \author Yannick de Hoop
+
+from dynamic_reconfigure.parameter_generator_catkin import *
```
jonbinney, line 37: Thanks for including dynamic_reconfigure for this!

include/laser_filters/angular_interval_filter.h
```diff
@@ -0,0 +1,223 @@
...
+  bool configure()
+  {
+    ros::NodeHandle private_nh("~" + getName());
+    dyn_server_.reset(new dynamic_reconfigure::Server<laser_filters::AngularIntervalFilterConfig>(own_mutex_, private_nh));
+    dynamic_reconfigure::Server<laser_filters::AngularIntervalFilterConfig>::CallbackType f;
+    f = boost::bind(&laser_filters::LaserScanAngularIntervalFilter::reconfigureCB, this, _1, _2);
+    dyn_server_->setCallback(f);
+
+
+    if (!filters::FilterBase<sensor_msgs::LaserScan>::getParam(std::string("interval_sets"), interval_sets_))
```
jonbinney, line 71: Doesn't `dynamic_reconfigure::Server` call your callback with default values when you instantiate it? I think it will read the parameter here so you could skip this part. I guess the one downside is that if the user doesn't set a ROS param to override the default values from the `.cfg` file, you won't know to warn them. But that's usually how dynamically reconfigureable params work in ros, so probably fine?

cfg/AngularIntervalFilter.cfg
jonbinney: I think the most reasonable default here should be probably be an empty list.

include/laser_filters/angular_interval_filter.h
```diff
@@ -0,0 +1,223 @@
...
+            filtered_scan.ranges[i] = replacement_value;
+        }
+      }
+    }
+    return true;
+  }
+  
+  void reconfigureCB(AngularIntervalFilterConfig& config, uint32_t level)
+  {
+  	interval_sets_ = config.interval_sets; // th11,th12;th21,th22;th31,th32
```
jonbinney, line 110: Does the interval_sets_ string need to be a member variable? It seems like you could parse it here and only store the `intervals_` as doubles.

jonbinney: Also it looks like the format in the comment doesn't match what is in the `.cfg` file. Shouldn't this be `[[th11 th12],[th3, th21]...`?

include/laser_filters/angular_interval_filter.h
```diff
@@ -0,0 +1,223 @@
...
+    std::string parse_error;
+    if(parseIntervalString(text, parse_error, intervals_))
+    {
+      ROS_INFO("Got intervals %s", parse_error.c_str());
+      got_intervals_ = true;
+    }
+    else
+    {
+      got_intervals_ = false;
+      ROS_WARN("Failed to calculate intervals: %s", parse_error.c_str());
```
jonbinney, line 123: I think it would be cleaner here to throw an exception on an error immediately, rather than printing out a warning. (This is what polygon filter does on a bad reconfigure parameter: https://github.com/ros-perception/laser_filters/blob/353392eb6d154312dd47c4d477f6324ae5782cdc/src/polygon_filter.cpp#L94 )

include/laser_filters/angular_interval_filter.h
```diff
@@ -0,0 +1,223 @@
...
+  {
+
+  }
+
+  bool update(const sensor_msgs::LaserScan& input_scan, sensor_msgs::LaserScan& filtered_scan)
+  {
+    filtered_scan = input_scan;
+    double replacement_value = std::numeric_limits<double>::quiet_NaN();
+    double current_angle = input_scan.angle_min;
+    if(got_intervals_ && enable_)
```
jonbinney, line 93: Consider making the name of this variable a bit more detailed. Maybe `have_valid_intervals`?

include/laser_filters/angular_interval_filter.h
```diff
@@ -0,0 +1,223 @@
...
+    }
+  }
+
+  static bool compare2dVect(std::vector<double> v1, std::vector<double> v2) 
+  { 
+    return (v1[0] < v2[0]); 
+  } 
+
+
+  bool parseIntervalString(const std::string& input, std::string& error_return, std::vector<std::vector<double>>& interval)
```
jonbinney, line 133: This function is a bit verbose, but it looks since it has worked in polygon filter, we can leave it alone.

laser_filters_plugins.xml
```diff
@@ -18,6 +18,12 @@
 	This is a filter which filters sensor_msgs::LaserScan messages based on range
       </description>
     </class>
+    <class name="laser_filters/LaserScanAngularIntervalFilter" type="laser_filters::LaserScanAngularIntervalFilter" 
+      base_class_type="filters::FilterBase<sensor_msgs::LaserScan>">
+      <description>
+  This is a filter which filters sensor_msgs::LaserScan messages based on angular intervals.
```
jonbinney, line 24: Could you add a couple more sentences to the description? Maybe something like `For example, if the intervals parameter is ...., then the filter will only allow laser beams whose angle is between .... and ....`

jonbinney (changes requested):
@ulassbin sorry for the long delay. This filter looks very useful. I've suggested some changes in the code; could you also fix the conflict in the cmakelists?

PR #111, opened 2021-04-09
Added DynamicReconfigure for RangeFilter

EricWiener:
Tried to make as few changes as possible (leaving style as is and
leaving implementation in the header file).

Also tried to make the code backward compatible.

If you want, I'd also be happy to clean up the styling a bit and move the implementation into a .cpp file, but didn't want to do it without checking first.


jonbinney:
Looks good; thanks @EricWiener !

PR #114, opened 2021-06-11
scan_to_cloud_filter_chain: Make cloud channels configurable

peci1:
Allow to configure channels of the resulting point cloud.

The defaults are set so that the node is backwards-compatible.

This is a revival of #74.


jonbinney (approved):

jonbinney:
@peci1 thanks for retargeting this to kinetic-devel. The CI is failing because of some docker rate issue, but tests pass locally and the code looks good. Merging.

PR #118, opened 2021-06-21
fix(speckle_filter): Possible segfault when ranges size was smaller than filter window

reinzor:


jonbinney (approved):
LGTM!

jonbinney:
Thanks @reinzor !

PR #119, opened 2021-07-02
Remove pointcloud footprint filter

jonbinney:
It has been deprecated for years, and is the only filter that depends on
pcl_ros. Removing it means we don't have to install the 500MB of
dependencies that pcl brings in on CI.


jonbinney:
If this causes issues for anyone, let me know. I think most cases should be solvable with the non-pointcloud footprint_filter.

PR #120, opened 2021-07-02
Enable CI on ros2 branch

jonbinney:
I expect the tests themselves to fail for now; I'll fix them in a later PR.


package.xml
```diff
@@ -20,8 +20,6 @@
   <depend>tf2_ros</depend>
   <depend>laser_geometry</depend>
   <depend>angles</depend>
```
jonbinney, line 22: rostime doesn't seem to be a package anymore, and we no longer need pcl_ros.

PR #121, opened 2021-07-02
Fix branch name for ros2 tests

jonbinney:


jonbinney:
Build fails in CI because filters fix https://github.com/ros/filters/pull/50 hasn't been released yet.

PR #123, opened 2021-07-15
Fix small bugs in build on galactic

jonbinney:
Tests must explicitly depend on rclcpp and the footprint filter was
missing an include.


jonbinney:
Progress! Remaining test failure should be fixed once filters is released again.

PR #126, opened 2021-08-04
Use filters package from debs in CI

jonbinney:
Using it from source was a temporary workaround until some upstream
fixes were released.


jonbinney:
This fails now because the new debs for the filters package haven't been synced to the public ROS apt repos yet. I'll re-run tests after that happens.

PR #127, opened 2021-09-27
change access specifiers for wrappers

renan028:
changing access specifier to `protected` (instead of `private`) so we can derive from base and change parameters dynamically with wrappers

### DIFF
--- a/include/laser_filters/box_filter.h
+++ b/include/laser_filters/box_filter.h
@@ -72,7 +72,7 @@ class LaserScanBoxFilter : public filters::FilterBase<sensor_msgs::LaserScan>
       const sensor_msgs::LaserScan& input_scan,
       sensor_msgs::LaserScan& filtered_scan);
 
-  private:
+  protected:
     bool inBox(tf::Point &point);
     std::string box_frame_;
     laser_geometry::LaserProjection projector_;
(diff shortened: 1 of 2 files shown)


jonbinney:
@renan028 what rosdistro are you on? Indigo-devel is pretty old

jonbinney:
Ooops nevermind... i don't know my alphabet. Indigo is the current ros1 branch. This change looks good to me. Merging.

jonbinney:
Wait I was right the first time.... kinetic-devel is the current ros1 branch. Are you still running indigo? Or jade?

renan028:
> Wait I was right the first time.... kinetic-devel is the current ros1 branch. Are you still running indigo? Or jade?

I am using melodic.
But if you check: https://index.ros.org/search/?term=laser_filters
And check the VCS version: https://index.ros.org/r/laser_filters/github-ros-perception-laser_filters/
It is indigo-devel

Edit: ok, just see that noetic is kinect-devel. Maybe melodic is just wrong. It would be nice to have the changes in kinect as well then.

jonbinney:
You're right.... why did i make a kinetic devel branch? Huh.

jonbinney:
I guess I could just rename kinetic-devel to noetic-devel and things would make sense.

PR #128, opened 2021-10-04
Fix compiler warnings

erwinbonsmatopic:
Fix all compiler warnings.

Fixed following warnings concerning filters header file includes:
`warning Including header <filters/filter_base.h> is deprecated, include <filters/filter_base.hpp> instead. [-Wcpp]
`
Also fixed format warning in speckle_filter.cpp:
`warning: format ‘%i’ expects argument of type ‘int’, but argument 8 has type ‘std::vector<float>::size_type’ {aka ‘long unsigned int’} [-Wformat=]`

Btw, this is the first in a series of intended PRs. The following ones will include additional unit tests and performance improvements.

### DIFF
--- a/include/laser_filters/angular_bounds_filter.h
+++ b/include/laser_filters/angular_bounds_filter.h
@@ -37,7 +37,7 @@
 #ifndef LASER_SCAN_ANGULAR_BOUNDS_FILTER_H
 #define LASER_SCAN_ANGULAR_BOUNDS_FILTER_H
 
-#include <filters/filter_base.h>
+#include <filters/filter_base.hpp>
 #include <sensor_msgs/LaserScan.h>
 
 namespace laser_filters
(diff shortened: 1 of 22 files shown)


JaccovdS:
@jonbinney ping

jonbinney:
@JaccovdS looks like the compiler isn't finding the headers in CI - ideas?

jonbinney:
Oh..... i think the headers didn't change until noetic. I've been resisting creating a noetic-devel branch, but that might be needed here.

jonbinney:
Well it seems that kinetic and melodic actually use the indigo-devel branch. So I think I can just rename the kinetic-devel branch to noetic-devel, and then release into noetic from that. Let me think it over.

erwinbonsmatopic:
@jonbinney, getting rid of compiler warnings is good practice, so that's why I did so. However, if these changes have too much impact on your CI and release set-up, we can park this PR. The other PRs do not depend on this. On the other hand, it may be good to properly support noetic. It's your call.

jonbinney:
It's worth getting this merged. I created a net noetic-devel branch and I'm cleaning up CI here: https://github.com/ros-perception/laser_filters/pull/138
Once that gets merged I'll retarget this PR to the new branch and merge it.

PR #129, opened 2021-10-04
Add speckle filter unit tests

erwinbonsmatopic:
Added google-test based unit tests for the speckle filter.

This prepares for refactoring the implementation to speed up performance.

The test include a very basic performance set-up, which executes the update method 10000 times. This can be used to compare the performance before/after refactoring. It is disabled by default. It can be enabled by passing the CATKIN_ENABLE_PERFORMANCE_TESTING to cmake.

Added configure method to LaserScanSpeckleFilter to enable unit testing filter without relying on ROS configuration infrastructure.

The new tests are executed when running ci.sh, they all pass and run in 0.09 seconds (on my machine)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,7 +70,15 @@ target_link_libraries(generic_laser_filter_node ${catkin_LIBRARIES} ${Boost_LIBR
 add_dependencies(laser_scan_filters ${PROJECT_NAME}_gencfg)
 
 if (CATKIN_ENABLE_TESTING)
+  if (CATKIN_ENABLE_PERFORMANCE_TESTING)
+    message("Enabling performance tests")
+    add_definitions(-DENABLE_PERFORMANCE)
+  else()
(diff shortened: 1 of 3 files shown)


JaccovdS:
@jonbinney ping

jonbinney:
This change looks good, thanks @JaccovdS ! I'm going to merge it into indigo-devel (from which melodic gets released) and the new noeticv-devel branch (from which noetic will be released from now on)

jonbinney:
Actually, retargeting this to indigo-devel got ugly; I'm going to merge this into  just noetic-devel for now. If you want it in melodic, go ahead and open a separate PR to the indigo-devel branch.

JaccovdS:
> This change looks good, thanks @JaccovdS ! I'm going to merge it into indigo-devel (from which melodic gets released) and the new noeticv-devel branch (from which noetic will be released from now on)

All credits to @erwinbonsmatopic I only did the ping ;)

PR #130, opened 2021-10-04
Speed up speckle filter implementation

erwinbonsmatopic:
These changes improve performance by a factor three.

Summary of changes:
- Do not allocate dynamic memory in each update invocation
- Speed up loop end-condition checks
- Algorithm improvements:
  - Perform out-of-range check during initialisation, only once for each
    sample
- Remove out of bound check from distance window validator.
  Instead, handle this by setting loop end criterion.

### DIFF
--- a/src/speckle_filter.cpp
+++ b/src/speckle_filter.cpp
@@ -73,8 +73,11 @@ bool LaserScanSpeckleFilter::configure()
 
 bool LaserScanSpeckleFilter::update(const sensor_msgs::LaserScan& input_scan, sensor_msgs::LaserScan& output_scan)
 {
+  boost::recursive_mutex::scoped_lock lock(own_mutex_);
+
   output_scan = input_scan;
-  std::vector<bool> valid_ranges(output_scan.ranges.size(), false);
+
+  std::vector<bool> &valid_ranges = valid_ranges_work_;
 
   /*Check if range size is big enough to use the filter window */
   if (output_scan.ranges.size() <= config_.filter_window + 1)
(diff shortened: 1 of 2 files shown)


JaccovdS:
@jonbinney ping

jonbinney:
@Patrick-Lascombe or @reinzor do either of you have time to review this code? I'm not very familiar with the speckle filter.

Patrick-Lascombe:
From what I can see the behavior is identical, but since I'm no longer using ROS1 I couldn't test it. Maybe @reinzor or someone else can test it more thoroughly.

src/speckle_filter.cpp
```diff
@@ -73,8 +73,11 @@ bool LaserScanSpeckleFilter::configure()
 
 bool LaserScanSpeckleFilter::update(const sensor_msgs::LaserScan& input_scan, sensor_msgs::LaserScan& output_scan)
 {
+  boost::recursive_mutex::scoped_lock lock(own_mutex_);
+
   output_scan = input_scan;
-  std::vector<bool> valid_ranges(output_scan.ranges.size(), false);
+
+  std::vector<bool> &valid_ranges = valid_ranges_work_;
```
Patrick-Lascombe, line 80: The only remark I have concern this line. Why copy a never instantiated global vector into valid_ranges, just to clear valid_ranges a few lines after. Wouldn't it be better to have valid_ranges as a global vector and clear it at each update ?

reinzor:
I will try to test it in our set-up today

reinzor:
Seems to work fine in simulation; today we will perform some more tests on our hardware setup

src/speckle_filter.cpp
reinzor: @erwinbonsmatopic , could you comment on this?

YannickdeHoop:
Seems to work fine with an actual hardware setup

src/speckle_filter.cpp
erwinbonsmatopic: Yes, apologies for the delay, I was offline for a week.

The local variable valid_ranges is a reference to the member variable, so line 80 does not perform any copying (I expect the compiler to not generate any code here). It's effectively letting the remainder of the code use the member variable without renaming the variable where it is used.

So in the first update invocation, dynamically memory is allocated as required when the vector is populated. As the local variable is a reference, the pointer to this memory is stored and retained in the member variable.

So the revised code avoids that each update needs to allocate dynamic memory. As long as the input range does not get larger between subsequent invocations, the update method re-uses the previously allocated memory.

PR #137, opened 2021-10-23
change_access specifier kinect

renan028:


jonbinney:
Thanks for this @renan028 !

PR #139, opened 2021-11-05
Split shadow filter headers

erwinbonsmatopic:
The code for the shadow filter had its entire implementation in its header file.

There not seem to be a good reason for this.
It is inconsistent with the implementation of the other filters.
It also slows down incremental builds.
It can also result it bloated binaries (as code may be built multiple times).

So this PR moves the implementation to cpp files.
Also made minor changes to header comments to improve consistency.
No changes have been made the filter's implementation. (This will be done in follow-up PRs)


jonbinney:
@erwinbonsmatopic looks like tests failed?

erwinbonsmatopic:
The problem is that in the CI build the tests do not link to the built library in `build/devel/lib` but instead to the installed library in `/opt/ros/noetic` (which is included in the ros:noetic-perception docker image).

A workaround is to adapt `ci.sh` to invoke cmake twice (directly after the first invocation). When doing this, the second invocation modifies `build/catkin_generated/setup_cached.sh` and adds the following line:
`export LD_LIBRARY_PATH="/home/erwin/views/laser_filters/build/devel/lib:$LD_LIBRARY_PATH"`

With this, the tests correctly link to the latest built code and pass.
Admittedly this is quite an ugly workaround.
Ideally the first cmake invocation already adds this required path but I am not sure yet how to achieve that.

Note the wrong-linking problem was already present before this changeset. It for example also occurs for the speckle filter tests. This might even explain why the shadow filter header files included the implementation. That may have been an even uglier workaround for this issue.

erwinbonsmatopic:
I added the CI workaround to this PR. Although not the perfect solution, this may be sufficient to enable the merge for several reasons. First, the problem was not introduced by this PR, it only reveals the existing problem. Second, this fixes the issue that the tests were not actually testing the built laser_filters library, but the installed library instead so that at least the tests can detect regressions. Third, it enables creation and review of PRs for subsequent improvements to the filter.

jonbinney:
Since directly using cmake was causing this problem, I've updated CI to use catkin_make: https://github.com/ros-perception/laser_filters/pull/140

Could you rebase this PR on noetic-devel and remove your changes to `ci.sh`?

erwinbonsmatopic:
Thanks for properly fixing ci.sh script. I rebased from noetic-devel (via a merge) and thereby undid my changes to ci.sh.

include/laser_filters/scan_shadow_detector.h
```diff
@@ -1,36 +1,36 @@
...
-*  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
-*  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
-*  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
-*  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
-*  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
-*  POSSIBILITY OF SUCH DAMAGE.
-*********************************************************************/
+ * Software License Agreement (BSD License)
+ * 
+ *  Copyright (c) 2017-2021, laser_filters authors
```
jonbinney: Other than the date, it looks like spaces were added, making the diff large. Could you revert the whitespace changes?

include/laser_filters/scan_shadows_filter.h
```diff
@@ -1,6 +1,7 @@
 /*
- * Copyright (c) 2008 Radu Bogdan Rusu <rusu@cs.tum.edu>
+ * Software License Agreement (BSD License)
```
jonbinney, line 2: Let's leave radu in the copyright line explicitly. How about `Copyright (c) 2008 Radu Bogdan Rusu <rusu@cs.tum.edu> and other laser_filters authors`

src/scan_shadow_detector.cpp
```diff
@@ -0,0 +1,75 @@
+/*********************************************************************
+ * Software License Agreement (BSD License)
+ * 
+ *  Copyright (c) 2017-2021, laser_filters authors
```
jonbinney: Same comment here - let's leave radu explicitly since he did the initial implementation, and add "other authors" in addition.

src/scan_shadows_filter.cpp
```diff
@@ -0,0 +1,169 @@
+/*
+/* Software License Agreement (BSD License)
+ *
+ * Copyright (c) 2008-2021, laser_filters authors
```
jonbinney: Same comments - leave radu in the license line

src/scan_shadows_filter.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+    ROS_DEBUG("ScanShadowsFilter removing %d Points from scan with min angle: %.2f, max angle: %.2f, neighbors: %d, and window: %d",
+              (int)indices_to_delete.size(), min_angle_, max_angle_, neighbors_, window_);
+    for (std::set<int>::iterator it = indices_to_delete.begin(); it != indices_to_delete.end(); ++it)
+    {
+      scan_out.ranges[*it] = std::numeric_limits<float>::quiet_NaN();  // Failed test to set the ranges to invalid value
+    }
+    return true;
+}
+}
```
jonbinney: Could you add an endline here?

jonbinney (changes requested):

include/laser_filters/scan_shadow_detector.h
erwinbonsmatopic: Done

include/laser_filters/scan_shadows_filter.h
erwinbonsmatopic: Okay, done

src/scan_shadow_detector.cpp
erwinbonsmatopic: Radu's name was not in original header file of scan_shadow_detector.h. In this file only laser_filter authors were mentioned in header comments. Underneath, Atsushi Watanabe was mentioned as author in separate comment with \author annotation.

So for new cpp file I just copied the original header from scan_shadow_detector.h

src/scan_shadows_filter.cpp
erwinbonsmatopic: Done

src/scan_shadows_filter.cpp
erwinbonsmatopic: Done

erwinbonsmatopic:
@jonbinney Thanks for reviewing. I handled your comments except for one (see individual responses for details).

jonbinney (approved):

jonbinney:
Thanks @erwinbonsmatopic, looks good!

PR #140, opened 2021-11-09
Use catkin_make to build and run tests on CI

jonbinney:
Using cmake directly wasn't handling the LD_LIBRARY_PATH correctly when
running tests, causing them to use the version of the laser_filters
library from /opt/ros/... instead of the one compiled from source.


jonbinney:
The problem this solves came up in https://github.com/ros-perception/laser_filters/pull/139

PR #141, opened 2021-11-12
Create shadow filter tests

erwinbonsmatopic:
We want to speed up the shadow filter.
However, as there were no tests yet, this PR adds those in preparation of that.

These tests test (and document) the current logic of the filter.

The filter logic seems a bit primitive in how it recognizes shadows.
Then again, the input data in the unit tests is also not realistic, so I assume it works fine in practice.

This set of tests is just a starting point to guard against regressions.
If others want to, they can add more realistic tests as needed.


jonbinney (approved):
Awesome, thanks @erwinbonsmatopic !

PR #144, opened 2021-11-17
Speed up shadow filter

erwinbonsmatopic:
Improve performance of scan shadow filter.

Changes:
- Instead of calculating sine/cosine each update, store (few) needed values in look-up tables and re-use those
- Remove use of set collection (and associated dynamic memory allocations) in update method
- Minor improvements to for-loop end criteria and checks inside loops

Together this results in speed-up of update method by more than factor three.

Added test to verify that lookup arrays is updated when needed.
Updated detector test to include negative angles (and more generally, match usage of detector in filter).

### DIFF
--- a/include/laser_filters/scan_shadow_detector.h
+++ b/include/laser_filters/scan_shadow_detector.h
@@ -46,18 +46,15 @@ namespace laser_filters
 class ScanShadowDetector
 {
 public:
+  friend class ScanShadowsFilter;
   float min_angle_tan_, max_angle_tan_;  // Filter angle thresholds
 
-  void configure(const float min_angle, const float max_angle, const int window);
-  void prepareForInput(const float angle_increment);
-  bool isShadow(const float r1, const float r2, const int angle_index);
+  void configure(const float min_angle, const float max_angle);
+  bool isShadow(const float r1, const float r2, const float included_angle);
 
   ~ScanShadowDetector();
 private:
-  int window_;
-  float angle_increment_;
-  std::vector<float> sin_map_;
-  std::vector<float> cos_map_;
+  bool isShadow(float r1, float r2, float included_angle_sin, float included_angle_cos);
 };
 }
 
--- a/include/laser_filters/scan_shadow_detector.h
+++ b/include/laser_filters/scan_shadow_detector.h
@@ -39,6 +39,8 @@
 #ifndef SCAN_SHADOW_DETECTOR_H
 #define SCAN_SHADOW_DETECTOR_H
 
+#include <vector>
+
 namespace laser_filters
 {
 class ScanShadowDetector
@@ -54,10 +56,8 @@ public:
 private:
   int window_;
   float angle_increment_;
-  float *sin_map_ = nullptr;
-  float *cos_map_ = nullptr;
-  float *shifted_sin_map_;
-  float *shifted_cos_map_;
+  std::vector<float> sin_map_;
+  std::vector<float> cos_map_;
 };
 }
 
--- a/include/laser_filters/scan_shadow_detector.h
+++ b/include/laser_filters/scan_shadow_detector.h
@@ -39,8 +39,6 @@
 #ifndef SCAN_SHADOW_DETECTOR_H
 #define SCAN_SHADOW_DETECTOR_H
 
-#include <cstddef>
-
 namespace laser_filters
 {
 class ScanShadowDetector
@@ -49,18 +47,17 @@ public:
   float min_angle_tan_, max_angle_tan_;  // Filter angle thresholds
 
   void configure(const float min_angle, const float max_angle, const int window);
-  bool isShadow(const float r1, const float r2, const int angle_index, const float angle_increment);
+  void prepareForInput(const float angle_increment);
+  bool isShadow(const float r1, const float r2, const int angle_index);
 
   ~ScanShadowDetector();
 private:
   int window_;
   float angle_increment_;
-  float *sin_map_ = NULL;
-  float *cos_map_ = NULL;
+  float *sin_map_ = nullptr;
+  float *cos_map_ = nullptr;
   float *shifted_sin_map_;
   float *shifted_cos_map_;
-
-  void updateAngleMap(const float angle_increment);
 };
 }
 
--- a/include/laser_filters/scan_shadow_detector.h
+++ b/include/laser_filters/scan_shadow_detector.h
@@ -46,8 +46,19 @@ class ScanShadowDetector
 public:
   float min_angle_tan_, max_angle_tan_;  // Filter angle thresholds
 
-  void configure(const float min_angle, const float max_angle);
-  bool isShadow(const float r1, const float r2, const float included_angle);
+  void configure(const float min_angle, const float max_angle, const int window);
+  bool isShadow(const float r1, const float r2, const int angle_index, const float angle_increment);
+
+  ~ScanShadowDetector();
+private:
+  int window_;
+  float angle_increment_;
+  float *sin_map_ = NULL;
+  float *cos_map_ = NULL;
+  float *shifted_sin_map_;
+  float *shifted_cos_map_;
+
+  void updateAngleMap(const float angle_increment);
 };
 }
 
--- a/include/laser_filters/scan_shadow_detector.h
+++ b/include/laser_filters/scan_shadow_detector.h
@@ -50,11 +50,25 @@ public:
   float min_angle_tan_, max_angle_tan_;  // Filter angle thresholds
 
   void configure(const float min_angle, const float max_angle);
+
+  /** \brief Check if the point is a shadow of another point within one laser scan.
+   * \param r1 the distance to the first point
+   * \param r2 the distance to the second point
+   * \param included_angle the angle between laser scans for these two points
+   */
   bool isShadow(const float r1, const float r2, const float included_angle);
 
-  ~ScanShadowDetector();
-private:
+  /** \brief Check if the point is a shadow of another point within one laser scan.
+   * Use this method instead of the version without the extra parameters to avoid
+   * computing sin and cos of the angle on every execution.
+   * \param r1 the distance to the first point
+   * \param r2 the distance to the second point
+   * \param included_angle_sin the sine of an angle between laser scans for these two points
+   * \param included_angle_cos the cosine of an angle between laser scans for these two points
+   */
   bool isShadow(float r1, float r2, float included_angle_sin, float included_angle_cos);
+
+  ~ScanShadowDetector();
 };
 }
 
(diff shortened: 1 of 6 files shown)


erwinbonsmatopic:
@jonbinney Can you have a look at this?

src/scan_shadows_filter.cpp
```diff
@@ -90,15 +90,16 @@ bool ScanShadowsFilter::configure()
       ROS_ERROR("max_angle must be 90 <= max_angle. Forcing max_angle = 90.\n");
       max_angle_ = 90.0;
     }
-    if (180 < min_angle_)
+    if (180 < max_angle_)
```
jonbinney, line 93: Was this just a bug in the code? Has it always been checking min_angle instead of min_angle?

src/scan_shadows_filter.cpp
```diff
@@ -129,42 +131,41 @@ bool ScanShadowsFilter::update(const sensor_msgs::LaserScan& scan_in, sensor_msg
     // copy across all data first
     scan_out = scan_in;
 
-    std::set<int> indices_to_delete;
+    int size = scan_in.ranges.size();
+    int maxY;
+    int maxNeighbors;
```
jonbinney: Could you use underscores instead of camel case? e.g. `max_y` and `max_neighbors`? I'm not sure that the convention is consistent in this package, but that's the ROS convention.

src/scan_shadow_detector.cpp
```diff
@@ -42,10 +42,21 @@
...
+    angle_increment_ = 0;
+    if (sin_map_ != nullptr)
+    {
+      delete [] sin_map_;
+      delete [] cos_map_;
+    }
+    sin_map_ = new float[window_ * 2 + 1];
+    cos_map_ = new float[window_ * 2 + 1];
+    shifted_sin_map_ = sin_map_ + window_;
+    shifted_cos_map_ = cos_map_ + window_;
```
jonbinney: Creating these shifted sin and cos arrays makes the code more complex (need to look here and where the array is actually used to understand what is happening). Would it make it that much slower to add `window_` in when we do the actual array lookup?

include/laser_filters/scan_shadow_detector.h
```diff
@@ -46,8 +46,18 @@ class ScanShadowDetector
...
-  bool isShadow(const float r1, const float r2, const float included_angle);
+  void configure(const float min_angle, const float max_angle, const int window);
+  void prepareForInput(const float angle_increment);
+  bool isShadow(const float r1, const float r2, const int angle_index);
+
+  ~ScanShadowDetector();
+private:
+  int window_;
+  float angle_increment_;
+  float *sin_map_ = nullptr;
```
jonbinney: These raw pointers worry me in terms of protecting against bugs in future patches. Why not use a `std::vector<float>` that will do the memory management for us?

src/scan_shadow_detector.cpp
```diff
@@ -42,10 +42,21 @@
...
   {
     min_angle_tan_ = tanf(min_angle);
     max_angle_tan_ = tanf(max_angle);
+    window_ = window;
+    angle_increment_ = 0;
+    if (sin_map_ != nullptr)
+    {
+      delete [] sin_map_;
+      delete [] cos_map_;
+    }
```
jonbinney: Do we need a lock on this configure method? This gets called by `ScanShadowsFilter::reconfigureCB()` which (if i remember correctly) gets called directly in the roscpp worker thread handling an incoming reconfigure request. If someone uses this filter in a node that uses `ros::AsyncSpinner()`, then we might get incoming scans to process after we've deleted the sin and cos maps, but before we regenerate them. What do you think of adding a mutex member variable and grabbing it with a `std::lock_guard` here and when the maps are used?

src/scan_shadow_detector.cpp
```diff
@@ -54,22 +65,35 @@ namespace laser_filters
       max_angle_tan_ = -max_angle_tan_;
   }
 
-  bool ScanShadowDetector::isShadow(const float r1, const float r2, const float included_angle)
-  {
-    const float perpendicular_y_ = r2 * sinf(included_angle);
-    const float perpendicular_x_ = r1 - r2 * cosf(included_angle);
-    const float perpendicular_tan_ = fabs(perpendicular_y_) / perpendicular_x_;
+  void ScanShadowDetector::prepareForInput(const float angle_increment) {
```
jonbinney: Having `prepareForInput()` be a separate function that the Filter class must remember to call seems like it could make it easy to add a bug later (someone doesn't know it has to be called). Why not call is every time in `ScanShadowDetector::isShadow()`?

jonbinney (changes requested):
Sorry for the slow response! This looks like a very useful PR. I've left some suggestions.

src/scan_shadows_filter.cpp
erwinbonsmatopic: This was a copy-paste bug in the original code. The code in the if-block was refering to the max_angle, but the check to the min_angle.

src/scan_shadows_filter.cpp
erwinbonsmatopic: Variables have been updated as suggested

include/laser_filters/scan_shadow_detector.h
erwinbonsmatopic: Using a vector here slows down the update loop by 25% on my machine. That's significant, but overall there is still a significant performance gain. I agree that pointers are riskier. I let it be your call if you want to switch. We have the code ready.

src/scan_shadow_detector.cpp
erwinbonsmatopic: I just tried, and the performance benefit here is negligible. If we stick with dynamically allocated arrays (see next comment), we can remove the use of the shifted maps.

src/scan_shadow_detector.cpp
erwinbonsmatopic: The maps are used in the isShadow method, which gets called (num_samples * window_size * 2) times per update, so adding a lock here is quite expensive. I would suggest letting the filter via its lock ensure that update and reconfigureCB never run concurrently. Do you agree?

src/scan_shadow_detector.cpp
erwinbonsmatopic: The drawback of moving it to the isShadow method is that it gets called (num_samples * window * 2) times per update instead of once. This results in about 5% regression in performance.

I propose to leave it as is. The risk of a regression is mitigated by the new unit test: SingleForwardShadow_AngleIncrementChanged. This test explicitly tests that the map is updated when needed and fails if this is not done.

include/laser_filters/scan_shadow_detector.h
jonbinney: I wonder why there is a performance loss at all - are you compiling in release mode? (`-DCMAKE_BUILD_TYPE=Release`?) The vector shouldn't be doing any bounds checking or anything in that case, so I don't see why it wouldn't compile down to the same machine code as the raw pointers.

src/scan_shadow_detector.cpp
jonbinney: Ah, I see. Before this PR, the `ScanShadowDetector` class only does the computation for a given pair of laser beams - it doesn't care how many beams are in the laser scan or how they are spaced. After the PR, `ScanShadowDetector` makes an assumption (which is true for all LaserScan messages) that there is a fixed number of beams in a scan, and that they are evenly spaced. 

I tend to be very conservative about changes to the class interfaces in this package. The code has been around for a long time, and is likely used in many different ways by different people. For example, nothing stops people from using the `ScanShadowDetector` class directly from their own code without using the filter node. That isn't ideal, but I expect that it does happen. Whenever possible on code as old and stable as this, I'd prefer to make changes that break other use cases as little as possible.

What do you think about putting the cached sin and cos maps in the `ScanShadowFilter` class, since it already has a concept of a "scan" rather than just dealing with individual beams like the `ScanShadowDetector` class? Then you could add a lower level function to `ScanShadowDetector` that skips the computation of sin and cos:

```c++

bool ScanShadowDetector::isShadow2(float r1, float r2, float included_angle_sin, float included_angle_cos)
{
  const float perpendicular_y_ = r2 * included_angle_sin;
  const float perpendicular_x_ = r1 - r2 * included_angle_cos;
  const float perpendicular_tan_ = fabs(perpendicular_y_) / perpendicular_x_;
  
  if (perpendicular_tan_ > 0)
  {
    if (perpendicular_tan_ < min_angle_tan_)
      return true;
  }
  else
  {
    if (perpendicular_tan_ > max_angle_tan_)
      return true;
  }
  return false;
}

bool ScanShadowDetector::isShadow(const float r1, const float r2, const float included_angle)
{
  float included_angle_sin = sinf(included_angle);
  float included_angle_cos = cosf(included_angle);
  return isShadow2(r1, r2, included_angle_sin, included_angle_cos)
}
```

Then the `ScanShadowFilter` class can handle caching the cos and sin values and call `ScanShadowDetector::isShadow2` to avoid doing the trig functions every time.

If i were writing this code from scratch, I would probably get rid of the ScanShadowDetector class completely and replace it with functions. Oh well.

src/scan_shadow_detector.cpp
jonbinney: Yes, that makes sense.

include/laser_filters/scan_shadow_detector.h
erwinbonsmatopic: When building with release settings vectors still seem slightly slower, but only 10% which is negligible. All in all, the performance gain is more than a factor three. So the code has been changed to use vectors

src/scan_shadow_detector.cpp
erwinbonsmatopic: With the look-up tables moving from the detector to the filter (see your comment below), there's no need anymore for synchronization at the detector. But we added a mutex to the reconfigureCB method

src/scan_shadow_detector.cpp
erwinbonsmatopic: Okay, makes sense. We moved the look-up tables to the filter. The new isShadow method that takes the sin/cos values has become a private API in the detector, only for use by the filter.

jonbinney:
Thanks, this looks almost ready to merge! Two more requests:
- could you add comments before each of the `isShadow` methods in the header, describing when to use each?
- i think the use of private and friend class here is unneeded complexity, and others may want to use the new, faster method. Could you remove the private and friend stuff and make the new isShadow public?

erwinbonsmatopic:
> Thanks, this looks almost ready to merge! Two more requests:
> 
> * could you add comments before each of the `isShadow` methods in the header, describing when to use each?
> * i think the use of private and friend class here is unneeded complexity, and others may want to use the new, faster method. Could you remove the private and friend stuff and make the new isShadow public?

Code reworked as requested.

jonbinney:
Oops, sorry, one last thing to change: could you remove the empty desctructor for the detector class?

erwinbonsmatopic:
> Oops, sorry, one last thing to change: could you remove the empty desctructor for the detector class?

Done. Also removed the lingering friend declaration that was not needed anymore.

jonbinney (approved):

jonbinney:
Thanks for all your work on this - it is a big improvement in speed!

PR #145, opened 2021-12-01
Add static polygon filter

erwinbonsmatopic:
This adds a variant of the existing polygon filter.

The new filter assumes that the transform from the robot's base to the laser filter is static, i.e. the position and orientation of the laser filter is fixed. This is typically the case. This assumption allows significantly lower CPU usage. First, it does not need to continuously listen to the active transform topic anymore. This ensures that laser filter chains for which there is no laser filter data (e.g. because they are temporarily disabled or are optional and not installed) consume no CPU. On our set-up, each dormant chain was still consuming 1% of CPU, which adds up if you have multiple chains. Second, the transform calculations can be removed from the update loop, as they only need to be performed once, during configuration.

I am not sure if the behavior of the old filter is required. If not, the new filter could replace the old one.
However, in order not to introduce any breaking changes, the new static filter is added alongside the old implementation. A common base class is added for code reuse.

### DIFF
--- a/include/laser_filters/polygon_filter.h
+++ b/include/laser_filters/polygon_filter.h
@@ -43,7 +43,7 @@
 #ifndef POLYGON_FILTER_H
 #define POLYGON_FILTER_H
 
-#include <filters/filter_base.hpp>
+#include <filters/filter_base.h>
 
 #include <sensor_msgs/LaserScan.h>
 #include <sensor_msgs/point_cloud_conversion.h>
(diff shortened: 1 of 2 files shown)


jonbinney:
One question for context before I do a detailed review on this PR - why a new filter in stead of adding a new parameter "dynamic_frame" or something similar, and have it default to `true`? That would preserve the old behavior while adding the new feature, without needing to add another filter. (no need to change any code yet, i'm just curious)

erwinbonsmatopic:
> One question for context before I do a detailed review on this PR - why a new filter in stead of adding a new parameter "dynamic_frame" or something similar, and have it default to `true`? That would preserve the old behavior while adding the new feature, without needing to add another filter. (no need to change any code yet, i'm just curious)

This is mainly guided by the implementation. The state that both variants require is quite different. See the header file in this PR.

I can understand that you may want to hide this implementation detail from filter users, and make it appear a single filter with a boolean parameter. Even in this case, I would suggest keeping the implementations separate. To achieve this, the filter logic would move to two "filter logic" components derived from a shared base class (with an API to configure and update). Depending on the new parameter, the PolygonFilter would instantiate and use the appropriate implementation.

This would make the implementation slightly more complex. Let me know if you would prefer this approach. If so, we can refactor the code accordingly.

jonbinney (approved):

jonbinney:
Sounds reasonable. Thanks @erwinbonsmatopic !

PR #149, opened 2022-01-15
added example config and launch file for scan blob filter

JohnTGZ:


jonbinney:
Looks good. Thanks @JohnTGZ !

PR #154, opened 2022-02-15
Register static polygon filter as plugin

erwinbonsmatopic:
PR #145 added a new filter, the static polygon filter.

However, it could not yet be used in launch files as it was not yet registered as plugin.
This PR fixes that.


YannickdeHoop:
ping @jonbinney

jonbinney:
LGTM. Thanks @erwinbonsmatopic  and thanks for the reminder @YannickdeHoop !

PR #156, opened 2022-02-28
adding support for invert-parameter to select if points within or out…

NikolasE:
…side of box are kept


NikolasE:
@jonbinney I deleted my first fork, I quickly made a new one with the same changes.

jonbinney:
Thanks @NikolasE ! CI is failing for rolling, but that looks completed unrelated to this change. Somehow it is trying to build python itself....

PR #157, opened 2022-03-15
Consistently publish polygon

erwinbonsmatopic:
Unlike the old version, the static polygon filter did not yet publish the polygon. This PR fixes that.

In doing so, it also removes code duplication related to filter configuration. The same configure method is now shared by both sub-classes.


jonbinney (approved):
LGTM!

PR #162, opened 2022-06-24
Do not use "not" as it is not defined in MSVC

wolfv:
Additionally, the symbols are also not `dllexport`'ed for Windows/MSVC.

I've fixed that by adding `set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS 1)`, but maybe you prefer manual visibility annotations?


jonbinney:
Sorry for the delayed response, and thanks for the fix! As far as exporting symbols - using CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS sounds good to me.

PR #163, opened 2022-07-05
Setting STATUS lvl to msgs reg. performance tests

rickvanosch:
Default level of CMake's Message is None/Notice, these are forwarded to stderr, triggering a warning in the build when using catkin build.

I'm not sure what the desired behaviour was, but IMO both messages should be on STATUS level or at least the one where the performance test is enabled.


rickvanosch:
Gentle ping!

jonbinney (approved):

jonbinney:
Thanks @rickvanosch !

rickvanosch:
Thanks @jonbinney !

Issue #71, opened 2019-04-01
Using laser filter as a nodelet


Patrick-Lascombe:
Good morning,

I wish to use laser_filters/LaserScanAngularBoundsFilterInPlace as a nodelet inside the nodelet manager that manage my laser scan. Is it possible or did someone already tried something like this ?

Thanks for your help

PR #169, opened 2022-11-03
ROS1 Nodelet Updates

marip8:
This PR adds the following updates:
- Fixes topic remapping for the `ScanToCloudFilterChain` nodelet
    -  In the current implementation, any topic remapping applied to the nodelet is ignored because the `ScanToCloudFilterChain` class uses its own public node handle, not the public node handle provided by the nodelet which accounts for topic remapping
- Adds a nodelet implementation for the `ScanToScanFilterChain` class
    - This class was also updated to accept public and private node handles from the nodelet to account for topic remapping correctly

Addresses #71


jonbinney:
Looks like a nice improvement! I just need to do a code review before I merge.

jonbinney (approved):
Looks good.

PR #171, opened 2022-11-09
[ROS-O] various cleanup

v4hn:
all but the last patch are required to compile with a current debian environment and are backward compatible.
I threw in the last one to avoid a serious looking warning by clang that actually has no effect here.


jonbinney:
@v4hn what are the target linux distros and compiler versions for ROS-O? I don't see it in the target platforms REP

v4hn:
Sorry for the late reply. Please see https://github.com/ros-o/ros-o (and my linked talk if you didn't watch it) for what ROS-O is and isn't. It's no "regular ROS distribution".

There is currently no limit on more current supported targets. Everything that remains compatible with Ubuntu 20.04 / the noetic REP entry should be contributed back.

jonbinney:
@v4hn ah, that makes sense. Thank you for the work you're doing!

jonbinney (approved):
Doesn't look like it will break anything for noetic, so should be good.

PR #175, opened 2023-03-01
Added declaration of parameters

tuxnese:
Hi, I have added the declaration of the parameters, otherwise it is impossible to set the value of the parameter on Galactic.


jonbinney:
Looks good. Thanks @tuxnese !

tuxnese:
> Looks good. Thanks @tuxnese !

Thank you for accepting it!

PR #177, opened 2023-03-30
use class_list_macros.hpp instead of .h

lucasw:
The .h files have been deprecated for years


jonbinney:
LGTM. Thanks @lucasw !

Issue #184, opened 2023-08-09
Filtered times are offset from sensor times.


HansLarsen:
The filtered times are offset and stuck in a 1 second interval loop.
![TIMESTAMPS](https://github.com/ros-perception/laser_filters/assets/45482017/42613d39-f2c1-4697-b6f0-5f7493372e27)
Heres plot juggler, the red line is the input laser scans and the orange line is the output from the filter.

I'm on commit 65cbefd
The distribution is ROS2 humble, everything build from the latest sources, according to the build from source guide, on Ubuntu 18.04.

The delay is likewise present if you print them in the callback:
```
Callback
  void callback(const std::shared_ptr<const sensor_msgs::msg::LaserScan> & msg_in)
  {
    RCLCPP_INFO(
      nh_->get_logger(), "%d, %d", msg_in->header.stamp.sec, msg_in->header.stamp.nanosec);
    // Run the filter chain
    if (filter_chain_.update(*msg_in, msg_)) {
      //only publish result if filter succeeded
      output_pub_->publish(msg_);
      RCLCPP_INFO(nh_->get_logger(), "%d, %d", msg_.header.stamp.sec, msg_.header.stamp.nanosec);
    }
  }
```
`[scan_to_scan_filter_chain-1] [INFO] [1691585756.502811379] [sick_tim_5xx.scan_to_scan_filter_chain]: 1691579413, 906227588
[scan_to_scan_filter_chain-1] [INFO] [1691585756.504288207] [sick_tim_5xx.scan_to_scan_filter_chain]: 1691579392, 906227588
[scan_to_scan_filter_chain-1] [INFO] [1691585757.162848118] [sick_tim_5xx.scan_to_scan_filter_chain]: 1691579413, 972615242
[scan_to_scan_filter_chain-1] [INFO] [1691585757.164226961] [sick_tim_5xx.scan_to_scan_filter_chain]: 1691579392, 972615242
[scan_to_scan_filter_chain-1] [INFO] [1691585757.827929301] [sick_tim_5xx.scan_to_scan_filter_chain]: 1691579414, 39208173
[scan_to_scan_filter_chain-1] [INFO] [1691585757.829719778] [sick_tim_5xx.scan_to_scan_filter_chain]: 1691579392, 39208173
[scan_to_scan_filter_chain-1] [INFO] [1691585758.502889853] [sick_tim_5xx.scan_to_scan_filter_chain]: 1691579414, 105945587
[...]

PR #186, opened 2023-09-27
Fix time stamp issue for angular filter

JosefGst:
- Changed the time increment to nanoseconds. 
This also solved the time stamp issue #184 for the angular filter for me.
- Added angular filter example


jonbinney:
@JosefGst thank you for doing this! Reviewing now.

examples/angular_filter_example.launch.py
```diff
@@ -0,0 +1,18 @@
...
+        Node(
+            package="laser_filters",
+            executable="scan_to_scan_filter_chain",
+            parameters=[
+                PathJoinSubstitution([
+                    get_package_share_directory("laser_filters"),
+                    "examples", "angular_filter_example.yaml",
+                ])],
+        )
+    ])
```
jonbinney: nit: could you add an endline here?

examples/angular_filter_example.yaml
```diff
@@ -0,0 +1,8 @@
+scan_to_scan_filter_chain:
+  ros__parameters:
+    filter1:
+      name: angle
+      type: laser_filters/LaserScanAngularBoundsFilter
+      params:
+        lower_angle: -1.52
+        upper_angle: 1.52
```
jonbinney: Nit: add endline at end of file

include/laser_filters/angular_bounds_filter.h
```diff
@@ -79,7 +79,7 @@ namespace laser_filters
           if(start_angle < lower_angle_){
             start_angle += input_scan.angle_increment;
             current_angle += input_scan.angle_increment;
-            start_time.set__sec(start_time.sec + input_scan.time_increment);
```
jonbinney, line 82: The original code does seem wrong -  `start_time.sec` is an int32, `input_scan.time_increment` is a float32. So adding them should result  in a float32. But then `set__sec` expects an int, so it has to round that float, and it only sets the `sec` field in start_time.

include/laser_filters/angular_bounds_filter.h
```diff
@@ -79,7 +79,7 @@ namespace laser_filters
...
+            start_time.set__nanosec(start_time.nanosec + (input_scan.time_increment * 10e9)); // convert time increment to nanoseconds
```
jonbinney: I think there are still a couple of issues:
- there are 1e9 nanoseconds in a second, so you should multiply by 1e9, not 10e9, right?
- set_nanosec only updates the number of nanoseconds.  The Time message type only allows nanosec to be between 0 and 1e9, so if the scan starts near the end of a second and we remove a bunch of beams from the beginning, then we could end up with an invalid number of nanoseconds. Instead we would want to add `nanoseconds / 1e9` to start_time.sec and `nanoseconds % 1e9` to start_time.nanosec

Time math is always tricky to get just right, so if there's a way to do this using utility functions that handle the details for us, that would be ideal.  The duration type should handle converting a floating point number of seconds to a proper duration for us in its from_seconds method: https://docs.ros2.org/foxy/api/rclcpp/classrclcpp_1_1Duration.html#a39ce5aa6d0bb190d7d59857aa7a9af1a

How about:
```
start_time += Duration::from_seconds(input_scan.time_increment);
```

examples/angular_filter_example.launch.py
```diff
@@ -0,0 +1,18 @@
+from launch import LaunchDescription
```
jonbinney, line 1: Thank you for adding the example!

jonbinney (changes requested):

include/laser_filters/angular_bounds_filter.h
JosefGst: Thank you for your review and pointing out the remaining issues.
You're right, it should be 1e9.
I will try to use the time duration as you recommend.

JosefGst: Sorry, was on holliday and didn't had the hardware for testing. But the start time is now calculated with the Duration object :smile: 
I tried with the rplidar which publishes:
```
angle_min: -3.1241390705108643
angle_max: 3.1415927410125732
```
And the settings for the filter:
```
scan_to_scan_filter_chain:
  ros__parameters:
    filter1:
      name: angle
      type: laser_filters/LaserScanAngularBoundsFilter
      params:
        lower_angle: -1.52
        upper_angle: 1.52  
```

In the second case I tried to cut off more data at the beginning. The start time of the filtered data (red) is slightly more delayed.
```
scan_to_scan_filter_chain:
  ros__parameters:
    filter1:
      name: angle
      type: laser_filters/LaserScanAngularBoundsFilter
      params:
        lower_angle: 1.5
        upper_angle: 1.52  
```

It looks correct to me. But please proof it.

include/laser_filters/angular_bounds_filter.h
```diff
@@ -99,6 +100,8 @@ namespace laser_filters
 
           }
         }
+        start_time.sec += d.seconds();
+        start_time.nanosec += d.nanoseconds();
```
jonbinney: Does `start_time += d` not work? I think that handling seconds and nanoseconds automatically for you is one of the nice part about using Duration objects.

JosefGst: Unfortunately I couldn't make `start_time += d` work.
Let me know if you have any other advice to improve. I'm very thankfull for your guidance :)

jonbinney: Ah, I see the problem. ROS2 has different types for serializing timestamps than for computing with time. Try changing the earlier line from:

```builtin_interfaces::msg::Time start_time = input_scan.header.stamp;```

to

```rclcpp::Time start_time = input_scan.header.stamp;```

Then `start_time` will be an `rclcpp:::Time` type, and `start_time += d;` should work.

Thanks for working on this!

JosefGst: Great, thank you for your advice. Now the code can be formated back to very similar as the original was.

jonbinney:
Looks great! I'm going to let the CI tests run one more time and then I'll merge. Thanks @JosefGst !

jonbinney (approved):

PR #191, opened 2024-03-13
Fix polygon padding on filter load

geotsam:
**The issue:**
Setting the polygon filter parameters from a `.yaml` results in correct parameters loaded in the parameter server (and can be seen in `rqt_reconfigure`). However, the filter itself does not incorporate the padding, unless the `reconfigureCb()` is called at least once. This can also be seen in rViz by inspecting the Polygon messages.

**The solution:**
Calling `padPolygon()` once in `configure()` in order to correctly pad the polygon inside the node.

### DIFF
--- a/src/polygon_filter.cpp
+++ b/src/polygon_filter.cpp
@@ -51,17 +51,14 @@
 #include <vector>
 
 /** @brief Same as sign(x) but returns 0 if x is 0. */
-inline double sign0(double x)
-{
-  return x < 0.0 ? -1.0 : (x > 0.0 ? 1.0 : 0.0);
-}
+inline double sign0(double x) { return x < 0.0 ? -1.0 : (x > 0.0 ? 1.0 : 0.0); }
 
 void padPolygon(geometry_msgs::Polygon& polygon, double padding)
 {
   // pad polygon in place
   for (unsigned int i = 0; i < polygon.points.size(); i++)
   {
-    geometry_msgs::Point32& pt = polygon.points[ i ];
+    geometry_msgs::Point32& pt = polygon.points[i];
     pt.x += sign0(pt.x) * padding;
     pt.y += sign0(pt.y) * padding;
   }
@@ -70,7 +67,8 @@ void padPolygon(geometry_msgs::Polygon& polygon, double padding)
 double getNumberFromXMLRPC(XmlRpc::XmlRpcValue& value, const std::string& full_param_name)
 {
   // Make sure that the value we're looking at is either a double or an int.
-  if (value.getType() != XmlRpc::XmlRpcValue::TypeInt && value.getType() != XmlRpc::XmlRpcValue::TypeDouble)
+  if (value.getType() != XmlRpc::XmlRpcValue::TypeInt &&
+      value.getType() != XmlRpc::XmlRpcValue::TypeDouble)
   {
     std::string& value_string = value;
     ROS_FATAL("Values in the polygon specification (param %s) must be numbers. Found value %s.",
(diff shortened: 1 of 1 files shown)


jonbinney:
This looks reasonable. Before I merge I'm going to dig into the initialization code of DynamicReconfigure to make sure I understand in exactly what sequence things happen. One question so that I can reproduce your exact use case: When you say you're setting the parameters from a yaml file, are you loading the yaml file onto the parameter server in your roslaunch file? Is it the same roslaunch file you're using to launch the laser filters node?

geotsam:
Yes @jonbinney, the parameters are loaded from a launch file like this:

```
<node pkg="laser_filters" type="scan_to_scan_filter_chain" output="screen" name="laser_filter_rear">
    <remap from="scan" to="raw_scan_rear" />
    <remap from="scan_filtered" to="scan_rear" />
    <rosparam command="load" file="$(find my_robot_bringup)/config/laser_filters_rear.yaml" />
</node>
```

I'll also provide you with the yaml:

```
scan_filter_chain:
- name: footprint_filter
  type: laser_filters/LaserScanPolygonFilter
  params:
    polygon_frame: base_link
    polygon: [[0.16, 0.27], [0.16, -0.27], [-0.45, -0.27], [-0.45, 0.27]] 
    invert: false
    polygon_padding: 0.30
```

Setting a large `polygon_padding` (like 0.3) will make the issue quickly apparent :sweat_smile:

jonbinney:
The problem here seems to come from dynamic_reconfigure not supporting variable size arrays. The initial value for `polygon` that is defined in the YAML and loaded onto the parameter  is an array, which is supported by XmlRpc (which the is how ROS1 represents parameters). Since dynamic reconfigure can't handle arrays, it expects  `polygon` to be a string. To workaround this, `LaserScanPolygonFilterBase::configure()` loads the array representation from the `polygon` parameter, converts it to a string, and then overwrites the parameter with that string. This is really ugly.... but mostly works.

As @geotsam  points out, `padPolygon()` doesn't get called until dynamic reconfigure calls `LaserScanPolygonFilterBase::reconfigureCb()`. That actually happens when `dyn_server_->setCallback(f);` is called earlier in `configure()`, but at that point the polygon parameter hasn't yet been converted from an array to a string. I think that waiting to call `dyn_server_->setCallback(f);` until later in the code will be a cleaner fix; i'll add a review comment to that effect.

src/polygon_filter.cpp
```diff
@@ -274,6 +274,8 @@ bool LaserScanPolygonFilterBase::configure()
   param_config.invert = invert_filter_;
   dyn_server_->updateConfig(param_config);
 
+  padPolygon(polygon_, polygon_padding);
```
jonbinney: Instead of calling `padPolygon() ` here, could you move `setCallback(f)` from line 261 to here? By doing that, `reconfigureCB()` is called after the `polygon` parameter has manually been converted from an array to a string. Also could you add a comment saying something to the effect of "Calling setCallback(f) here calls `reconfigureCB()` which updates the polygon padding.

jonbinney (changes requested):

src/polygon_filter.cpp
geotsam: I tested your suggestion and it works, so I pushed a commit to do just that. I agree, that this is a cleaner and more reasonable way to resolve the issue.

geotsam:
Some auto-formatting slipped in the previous commit and modified the whole file  😖; I pushed a new one that only changes the lines we are discussing.

jonbinney:
One more nitpick - could you add back the end line at the end of the file? Looks like that was left in from the autoformat.

geotsam:
It should be fine now! I added the end line back.

jonbinney (approved):

jonbinney:
Thank you @geotsam!

geotsam:
Hi @jonbinney! Do you know if the apt noetic package is going to be updated to include this fix?

jonbinney:
@geotsam I'll do a release this week. Sorry it has taken a while!

jonbinney:
Tracking release process in this issue: https://github.com/ros-perception/laser_filters/issues/200

Issue #194, opened 2024-07-15
Add an option for lazy subscription on scan topic


bjsowa:
I use rplidar node with `auto_standby` option which stops the LIDAR motor when there is no subscription. Using it together with a laser filter forces the motor to constantly run as the filter chains constantly subscribes on the `scan` topic, even when there is no subscription on `scan_filtered` topic.

I think it would be beneficial to implement lazy subscription using rmw matched events, similar to how it's done in ros-perception/image_pipeline#815

PR #195, opened 2024-07-16
Add Lazy subscription

bjsowa:
Implements #194 .

this requires Matched publisher events, which are supported from ROS Iron onward, so a separate branch for ROS Humble is required.

This PR also fixes some issues with parameters in scan filter chains:
1. Explicitly declare all parameters , otherwise `get_parameter` won't work.
2. Remove unused parameters.
3. Add read-only flag to all parameters as the filter chains don't support dynamically changing them.

I also changed `spin_some` to `spin` as I see absolutely no reason to use the first method.

I can split this PR into multiple ones if needed,

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,10 @@ ament_auto_find_build_dependencies()
 # Build
 ##############################################################################
 
+if($ENV{ROS_DISTRO} STREQUAL "humble")
+    add_compile_definitions(IS_HUMBLE)
+endif()
+
 ament_auto_add_library(laser_scan_filters SHARED src/laser_scan_filters.cpp)
 ament_auto_add_library(laser_filter_chains SHARED
                          src/scan_to_cloud_filter_chain.cpp
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,8 +12,8 @@ ament_auto_find_build_dependencies()
 # Build
 ##############################################################################
 
-if($ENV{ROS_DISTRO} STREQUAL "humble")
-    add_compile_definitions(IS_HUMBLE)
+if(${rclcpp_VERSION_MAJOR} GREATER_EQUAL 20)
+    add_compile_definitions(RCLCPP_SUPPORTS_MATCHED_CALLBACKS)
 endif()
 
 ament_auto_add_library(laser_scan_filters SHARED src/laser_scan_filters.cpp)
--- a/package.xml
+++ b/package.xml
@@ -21,7 +21,6 @@
   <depend>rclcpp</depend>
   <depend>rclcpp_lifecycle</depend>
   <depend>rclcpp_components</depend>
-  <depend>ros_environment</depend>
   <depend>sensor_msgs</depend>
   <depend>tf2</depend>
   <depend>tf2_ros</depend>
--- a/package.xml
+++ b/package.xml
@@ -21,6 +21,7 @@
   <depend>rclcpp</depend>
   <depend>rclcpp_lifecycle</depend>
   <depend>rclcpp_components</depend>
+  <depend>ros_environment</depend>
   <depend>sensor_msgs</depend>
   <depend>tf2</depend>
   <depend>tf2_ros</depend>
--- a/src/scan_to_cloud_filter_chain.cpp
+++ b/src/scan_to_cloud_filter_chain.cpp
@@ -54,7 +54,7 @@ ScanToCloudFilterChain::ScanToCloudFilterChain(
   read_only_desc.read_only = true;
 
   // Declare parameters
-  #ifndef IS_HUMBLE
+  #ifdef RCLCPP_SUPPORTS_MATCHED_CALLBACKS
   this->declare_parameter("lazy_subscription", false, read_only_desc);
   #endif
   this->declare_parameter("high_fidelity", false, read_only_desc);
@@ -64,7 +64,7 @@ ScanToCloudFilterChain::ScanToCloudFilterChain(
   this->declare_parameter("laser_max_range", DBL_MAX, read_only_desc);
 
   // Get parameters
-  #ifndef IS_HUMBLE
+  #ifdef RCLCPP_SUPPORTS_MATCHED_CALLBACKS
   this->get_parameter("lazy_subscription", lazy_subscription_);
   #endif
   this->get_parameter("high_fidelity", high_fidelity_);
@@ -85,7 +85,7 @@ ScanToCloudFilterChain::ScanToCloudFilterChain(
     this->get_node_timers_interface());
   buffer_.setCreateTimerInterface(timer_interface);
 
-  #ifndef IS_HUMBLE
+  #ifdef RCLCPP_SUPPORTS_MATCHED_CALLBACKS
   if (lazy_subscription_) {
     rclcpp::PublisherOptions pub_options;
     pub_options.event_callbacks.matched_callback =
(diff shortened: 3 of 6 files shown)


jonbinney:
This seems like a good feature, thanks @bjsowa ! I've got a couple other PRs to review first, but hoping to get to this soon.

jonbinney:
Actually looks like https://github.com/ros-perception/laser_filters/pull/188 created some merge conflicts for this PR when I merged it. Could you fix those before I review the code?

bjsowa:
I manually reapplied the changes as there were too many conflicts.

The first 2 commits are not related to lazy subscription functionality. I can split them into separate PRs.

The third commit adds the lazy subscription and drops support for ROS Humble.

bjsowa:
I split this PR to only include the lazy subscription part. @jonbinney I would appreciate if you could look at  #197, #198 and #199 first

jonbinney:
Thanks for splitting this up, that will make it easier to review. Quick question - is it feasible to change this to avoid branching for humble (for example by using preprocessor macros?). As soon as we branch for humble we've got to start cherry picking bug fixes, which gets painful quickly.

bjsowa:
This should now build on Humble. I took some inspiration from [depthai-ros](https://github.com/luxonis/depthai-ros/blob/iron/depthai_bridge/CMakeLists.txt#L85-L91), though I'm now 100% sure it will work on the buildfarm

bjsowa:
@jonbinney  As all the "child" PR are merged, this one is ready for review. I squashed all commits as it was mostly resolving merge conflicts.

CMakeLists.txt
```diff
@@ -12,6 +12,10 @@ ament_auto_find_build_dependencies()
 # Build
 ##############################################################################
 
+if($ENV{ROS_DISTRO} STREQUAL "humble")
+    add_compile_definitions(IS_HUMBLE)
+endif()
```
jonbinney, line 17: This works, but I think what we actually want is to check the version of the rclcpp package. Cmake sets version variables after they are find_package'd; how about checking that '${rclcpp_VERSION_MAJOR}' is greater than or equal to 20? (it looks like matched callbacks were added in version 20)

package.xml
```diff
@@ -21,6 +21,7 @@
   <depend>rclcpp</depend>
   <depend>rclcpp_lifecycle</depend>
   <depend>rclcpp_components</depend>
+  <depend>ros_environment</depend>
```
jonbinney: I think we can skip this dependency if we use rclcpp_VERSION_MAJOR in the cmakelists.

CMakeLists.txt
jonbinney: Also it would be clearer to call the compile define RCLCPP_SUPPORTS_MATCHED_CALLBACKS

bjsowa: I was looking for a version macro in rclcpp package and couldn't find one. I was not aware of version variables set by cmake. I applied your suggestions.
btw. a lot of ROS packages don't use a consistent versioning scheme but rclcpp is a quality level 1 package so I have more hope.

src/scan_to_cloud_filter_chain.cpp
```diff
@@ -80,11 +85,30 @@ ScanToCloudFilterChain::ScanToCloudFilterChain(
...
+      [this](rclcpp::MatchedInfo & s)
+      {
+        if (s.current_count == 0) {
+          scan_sub_.unsubscribe();
+        } else if (!scan_sub_.getSubscriber()) {
+          scan_sub_.subscribe(this, "scan", rmw_qos_profile_sensor_data);
+        }
+      };
+    cloud_pub_ = this->create_publisher<sensor_msgs::msg::PointCloud2>("cloud_filtered",
+      rclcpp::SensorDataQoS(), pub_options);
```
jonbinney: It looks like SensorDataQoS sets the "depth" (number of messages to hang on to in case publishing gets backed up) to 5. If there's not a good reason to change it, I'd prefer to keep it to 10 (see the original create_publisher call).  There are also some other QoS settings that differ in SensorDataQoS from just what you get if you just pass in the depth as a number. I think that for this PR, it would be better not to change those. You could do so in a separate PR, but we'd want to go over each element of the QoS struct and discuss why it should be changed.

src/scan_to_cloud_filter_chain.cpp
```diff
@@ -80,11 +85,30 @@ ScanToCloudFilterChain::ScanToCloudFilterChain(
...
+  } else {
+    cloud_pub_ = this->create_publisher<sensor_msgs::msg::PointCloud2>("cloud_filtered",
```
jonbinney: Same thing - prefer not to change QoS in this PR unless there is a strong reason

src/scan_to_scan_filter_chain.cpp
```diff
@@ -80,8 +85,30 @@ ScanToScanFilterChain::ScanToScanFilterChain(
...
+    pub_options.event_callbacks.matched_callback =
+      [this](rclcpp::MatchedInfo & s)
+      {
+        if (s.current_count == 0) {
+          scan_sub_.unsubscribe();
+        } else if (!scan_sub_.getSubscriber()) {
+          scan_sub_.subscribe(this, "scan", rmw_qos_profile_sensor_data);
+        }
+      };
+    output_pub_ = this->create_publisher<sensor_msgs::msg::LaserScan>("scan_filtered",
```
jonbinney: Same thing  as in the scan_to_cloud_filter_chain.cpp - I'd prefer we just pass in 1000 as the QoS for this PR.

src/scan_to_scan_filter_chain.cpp
```diff
@@ -80,8 +85,30 @@ ScanToScanFilterChain::ScanToScanFilterChain(
...
+      rclcpp::SensorDataQoS(), pub_options);
+  } else {
+    output_pub_ = this->create_publisher<sensor_msgs::msg::LaserScan>("scan_filtered",
+      rclcpp::SensorDataQoS());
+    scan_sub_.subscribe(this, "scan", rmw_qos_profile_sensor_data);
+  }
+  #else
+  output_pub_ = this->create_publisher<sensor_msgs::msg::LaserScan>(
```
jonbinney: Same

CMakeLists.txt
jonbinney: True that there's no strict versioning scheme across ROS, but I think we can at least expect the major version not to _decrease_ :-)

jonbinney (changes requested):
Looks good except for making sure we keep the QoS settings the same. After that is fixed, this should be good to merge!

bjsowa:
@jonbinney  I kept the old QoS settings. A depth value of 1000 seems like an overkill (especially for a _reliable_ publisher) but I'll stick to changes only related to lazy subscription.

jonbinney (approved):

jonbinney:
Looks good! Yes, 1000 is a lot of scans to keep. I suspect the original thought was "lets keep as many as we can without using crazy amounts of memory". Typical 2D scanners may have around 1000 ranges and intensities per scan, which would be about `(1000 points) * (4 bytes for intensity + 4 bytes for range) = 8000 bytes per scan`. So 1000 scans would be around 8 MB, which would probably be fine even on most single board computers.

As far as reliability, even if the QoS is set to "reliable", if you are streaming data over wifi (say running SLAM on a desktop while a mobile robot drives around a building) you may lose connection here and there, especially when transitioning between access points.  It's nice not to lose those scans.

Thanks for this PR! And the others you made - they are big improvements!

PR #196, opened 2024-07-16
Update list of ros2 distros to test on

jonbinney:
Foxy and galactic are EOL; Iron is now supported.


bjsowa:
Why not add Jazzy as well?

jonbinney:
@bjsowa because i forgot! Thanks, adding now :-)

jonbinney:
Hooray, tests pass on the new distros!

PR #197, opened 2024-07-17
Don't use shared_from_this in the constructor

bjsowa:
Using shared_from_this inside a constructor is dangerous:

> If shared_from_this is called on an object that is not previously shared by [std::shared_ptr](https://en.cppreference.com/w/cpp/memory/shared_ptr), [std::bad_weak_ptr](https://en.cppreference.com/w/cpp/memory/bad_weak_ptr) is thrown (by the shared_ptr constructor from a default-constructed weak-this).

Right now, running `scan_to_cloud_filter_chain` results in `std::bad_weak_ptr` exception.

This PR fixes it by using a different constructor for `tf2_ros::MessageFilter`

### DIFF
--- a/src/scan_to_cloud_filter_chain.cpp
+++ b/src/scan_to_cloud_filter_chain.cpp
@@ -46,7 +46,8 @@ ScanToCloudFilterChain::ScanToCloudFilterChain(
   buffer_(this->get_clock()),
   tf_(buffer_),
   sub_(this, "scan", rmw_qos_profile_sensor_data),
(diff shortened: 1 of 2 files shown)


jonbinney (approved):
LGTM. Thanks for fixing this bug!! Can really tell people haven't been using that node and that our test coverage needs improvement :-/

bjsowa:
> LGTM. Thanks for fixing this bug!! Can really tell people haven't been using that node and that our test coverage needs improvement :-/

This bug was introduced by #188 so I caught it just a day after. I agree that tests should have detected it though.

PR #198, opened 2024-07-17
Create node executables using rclcpp_component macros

bjsowa:
`rclcpp_components_register_node` macro can create templated source for node executable that runs the Node component.

This PR replaces some boilerplate code using this macro.


jonbinney:
I like using the macros, but I think there's still a good reason to put them in separate `_node` source files. In https://github.com/ros-perception/laser_filters/pull/188 the main function was split into a separate file so that the filter chain component classes could be `#include`d by other people's code without getting a main function.

bjsowa:
> I like using the macros, but I think there's still a good reason to put them in separate `_node` source files.

Sure, If you feel so, you can close this PR. The macros just seem "cleaner" to me but it's just my opinion.

If you want to stick to `_node` source file, I will at least try to get 0597364 merged.

CMakeLists.txt
```diff
@@ -16,17 +16,18 @@ ament_auto_add_library(laser_scan_filters SHARED src/laser_scan_filters.cpp)
 ament_auto_add_library(laser_filter_chains SHARED
                          src/scan_to_cloud_filter_chain.cpp
                          src/scan_to_scan_filter_chain.cpp)
-rclcpp_components_register_nodes(laser_filter_chains
-                         "ScanToCloudFilterChain"
-                         "ScanToScanFilterChain")
 
-set(FILTER_CHAINS
-    scan_to_cloud_filter_chain
-    scan_to_scan_filter_chain
```
jonbinney, line 25: Could you add a comment describing what rclcpp_components_register_node does here? It's a very cool and useful macro, but the documentation isn't super easy to find, and it'd be nice to have a summary here.

jonbinney (changes requested):
I read up on rclcpp_components_register_node and now this PR makes sense to me. Once you add the suggested comment, I'll merge this.

CMakeLists.txt
bjsowa: I added a comment that briefly describes the macro

jonbinney (approved):

jonbinney:
Looks good, merging! btw i find it very odd that rclcpp_components_register_node creates the templated node executable, but rclcpp_components_register_nodes (plural) doesn't.....

PR #199, opened 2024-07-17
Sort up ROS parameters for filter chains

bjsowa:
Cherry picked from #195 

> * Explicitly declare all parameters , otherwise `get_parameter` won't work.
> * Remove unused parameters.
> * Add read-only flag to all parameters as the filter chains don't support dynamically changing them.


jonbinney (approved):
LGTM. Setting them all to read only makes sense - if we ever implement updating for them, we can change that.

jonbinney:
Thanks for this!

PR #201, opened 2024-08-06
Spatial median filter

YBachmann:
This PR adds a 1D spatial median filter. 
The size of the filter can be set via the parameter `window_size`. 

In this screenshot you can see the raw laserscan in blue and the result of the spatial median filter in red (`window_size=31`). 
The filter smoothes the walls and also removes isolated/scattered noise (see e.g. directly right of the robot). 
spatial_median_filter

### DIFF
--- a/include/laser_filters/median_spatial_filter.h
+++ b/include/laser_filters/median_spatial_filter.h
@@ -44,14 +44,18 @@
 #include <algorithm>
 #include <vector>
 #include <stdexcept>
+#include <limits>
+#include <cmath>
 
 namespace laser_filters
 {
 
 class LaserScanMedianSpatialFilter : public filters::FilterBase<sensor_msgs::msg::LaserScan>
 {
-public:
+private:
   int window_size_;
+
+public:
 
   bool configure()
   {
@@ -81,36 +85,66 @@ public:
     filtered_scan = input_scan;
 
     int half_window = window_size_ / 2;
-    std::vector<float> window;
+    std::vector<float> valid_values;
+    int nan_count = 0;
+    int neg_inf_count = 0;
+    int pos_inf_count = 0;
 
-    for (size_t i = 0; i < input_scan.ranges.size(); ++i)
+    for (size_t current_beam_index = 0; current_beam_index < input_scan.ranges.size(); ++current_beam_index)
     {
-      window.clear();
+        valid_values.clear();
+        nan_count = 0;
+        neg_inf_count = 0;
+        pos_inf_count = 0;
 
-      // Collect points within the window
-      for (int j = -half_window; j <= half_window; ++j)
-      {
-        int index = i + j;
-
-        if (index >= 0 && index < input_scan.ranges.size())
+        // Collect points within the window
+        for (int window_offset = -half_window; window_offset <= half_window; ++window_offset)
         {
-          if (!std::isnan(input_scan.ranges[index]))
-          {
-            window.push_back(input_scan.ranges[index]);
-          }
-        }
-      }
+            int index = current_beam_index + window_offset;
 
-      if (!window.empty())
-      {
-        // Calculate median
-        std::sort(window.begin(), window.end());
-        filtered_scan.ranges[i] = window[window.size() / 2];
-      }
-      else
-      {
-        filtered_scan.ranges[i] = std::numeric_limits<float>::quiet_NaN();
-      }
+            if (index >= 0 && index < input_scan.ranges.size())
+            {
+                float value = input_scan.ranges[index];
+
+                if (std::isnan(value))
+                {
+                    nan_count++;
+                }
+                else if (value == -std::numeric_limits<float>::infinity())
+                {
+                    neg_inf_count++;
+                }
+                else if (value == std::numeric_limits<float>::infinity())
+                {
+                    pos_inf_count++;
+                }
+                else
+                {
+                    valid_values.push_back(value);
+                }
+            }
+        }
+
+        // Determine which set is the largest
+        // In case of a tie, prioritize valid-values over nan-values over neg-inf-values over pos-inf-values
+        if (valid_values.size() >= nan_count && valid_values.size() >= neg_inf_count && valid_values.size() >= pos_inf_count)
+        {
+            // Sort the valid values and return the median
+            std::sort(valid_values.begin(), valid_values.end());
+            filtered_scan.ranges[current_beam_index] = valid_values[valid_values.size() / 2];
+        }
+        else if (nan_count >= valid_values.size() && nan_count >= neg_inf_count && nan_count >= pos_inf_count)
+        {
+            filtered_scan.ranges[current_beam_index] = std::numeric_limits<float>::quiet_NaN();
+        }
+        else if (neg_inf_count >= valid_values.size() && neg_inf_count >= nan_count && neg_inf_count >= pos_inf_count)
+        {
+            filtered_scan.ranges[current_beam_index] = -std::numeric_limits<float>::infinity();
+        }
+        else if (pos_inf_count >= valid_values.size() && pos_inf_count >= nan_count && pos_inf_count >= neg_inf_count)
+        {
+            filtered_scan.ranges[current_beam_index] = std::numeric_limits<float>::infinity();
+        }
     }
 
     return true;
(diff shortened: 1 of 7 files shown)


jonbinney:
This looks quite useful! I'll take a closer look tomorrow.

include/laser_filters/median_spatial_filter.h
```diff
@@ -0,0 +1,122 @@
...
+
+class LaserScanMedianSpatialFilter : public filters::FilterBase<sensor_msgs::msg::LaserScan>
+{
+public:
+  int window_size_;
+
+  bool configure()
+  {
+    // Default window size
+    window_size_ = 3;
```
jonbinney: Setting 3 here doesn't actually do anything - `FilterBase::getParam()` sets parameters to a default value equal to the default value of that C++ type. In this case that would be zero. I think it is actually good in this case to force users to set the "window_size" explicitly in the YAML. Here you don't need to initialize the `window_size_` variable at all (it will be initialized to zero when the class is constructed anyway)

include/laser_filters/median_spatial_filter.h
```diff
@@ -0,0 +1,122 @@
...
+#include <vector>
+#include <stdexcept>
+
+namespace laser_filters
+{
+
+class LaserScanMedianSpatialFilter : public filters::FilterBase<sensor_msgs::msg::LaserScan>
+{
+public:
+  int window_size_;
```
jonbinney, line 56: Any reason `window_size_` needs to be public? I realize that some other filters do this, but I think that it'd be best to make it private.

include/laser_filters/median_spatial_filter.h
```diff
@@ -0,0 +1,122 @@
...
+  bool configure()
+  {
+    // Default window size
+    window_size_ = 3;
+    getParam("window_size", window_size_);
+
+    // Ensure window size is positive
+    if (window_size_ <= 0)
+    {
+      throw std::runtime_error("Window size must be positive");
```
jonbinney: Instead of throwing an exception here, you can return false - the FilterChain will then know configuration failed.

include/laser_filters/median_spatial_filter.h
```diff
@@ -0,0 +1,122 @@
...
+    std::vector<float> window;
+
+    for (size_t i = 0; i < input_scan.ranges.size(); ++i)
+    {
+      window.clear();
+
+      // Collect points within the window
+      for (int j = -half_window; j <= half_window; ++j)
+      {
+        int index = i + j;
```
jonbinney: For readability consider more descriptive names for the indices. Maybe "window_center", "window_offset", and "beam_i" instead of "i", "j", and "index".

include/laser_filters/median_spatial_filter.h
```diff
@@ -0,0 +1,122 @@
...
+    // Ensure window size is positive
+    if (window_size_ <= 0)
+    {
+      throw std::runtime_error("Window size must be positive");
+    }
+
+    // Ensure window size is odd
+    if (window_size_ % 2 == 0)
+    {
+      window_size_ += 1;
```
jonbinney, line 76: Is it necessary to modify the window size here? Typically the median of a set with an even number of elements is equal to the average of the center two elements - people might find that behavior more expected.

include/laser_filters/median_spatial_filter.h
```diff
@@ -0,0 +1,122 @@
...
+      window.clear();
+
+      // Collect points within the window
+      for (int j = -half_window; j <= half_window; ++j)
+      {
+        int index = i + j;
+
+        if (index >= 0 && index < input_scan.ranges.size())
+        {
+          if (!std::isnan(input_scan.ranges[index]))
```
jonbinney: It makes sense that you can't find the median of an vector that includes NaN values, but what is the desired behavior in this case? If the window size is 101, and there are 100 NaNs in a row, do we really want to fill in all those NaNs with that one non-NaN value? (not saying this is wrong, but curious about the thought process here.

My initial thoughts:
- For obstacle avoidance, this could cause the robot to be overconfident if that one non-NaN value is large (it could end up clearing costmap cells thinking it got a long distance value for the ranges)
- For localization, this could cause the robot to think there is a big object or wall there, causing it to mislocalize.

laser_filters_plugins.xml
```diff
@@ -60,6 +60,12 @@
 	 This is a filter that removes points on directions defined in a mask from a laser scan.
       </description>
     </class>
+    <class name="laser_filters/LaserScanMedianSpatialFilter" type="laser_filters::LaserScanMedianSpatialFilter" 
+	    base_class_type="filters::FilterBase&lt;sensor_msgs::msg::LaserScan&gt;">
+      <description>
+	This is a spatial 1D median filter which filters sensor_msgs::msg::LaserScan messages.
```
jonbinney: Could you expand the description a bit to explain when this is useful? For example that it removes outliers and smooths, as you say in the PR description.

jonbinney (changes requested):
Added some questions. Also could you add an example yaml and python launch file in the examples folder?

include/laser_filters/median_spatial_filter.h
YBachmann: The reasoning behind this is that we always modify the current range (`filtered_scan.ranges[i]`) by looking at the ranges around it (e.g. the one directly left and directly right of it if `window_size_=3`). If we have an even window_size, e.g. 4, we would need to either look at two values to the right and one value to the left or the other way around. In other words: If the window_size is even, the current range, which will be modified cannot be centered in the window. 
I would propose to keep it this way but add a warning output if the window_size is even and we have to increment it.

jonbinney: Yes, your explanation makes sense. Sounds good to me.

include/laser_filters/median_spatial_filter.h
YBachmann: I think this is an important point :+1: 

My proposed solution would be to keep track of the number of NaN, +inf and -inf values and return such a value if it makes up the majority of the elements in the window. 
This way if the window mostly consists of valid values, we use the median of those. Otherwise we use the majority NaN/Inf-value.

jonbinney: That sounds like a good solution.

jonbinney:
It looks like you've made some changes to address my comments. When you have a chance, could you reply to each review comment to note that you've addressed it, and let me know when this PR is ready for another look?

include/laser_filters/median_spatial_filter.h
YBachmann: Removed redundant initializing of `window_size_`

include/laser_filters/median_spatial_filter.h
YBachmann: Print warning and return false instead of throwing exception.

include/laser_filters/median_spatial_filter.h
YBachmann: Changed names to something more descriptive.

laser_filters_plugins.xml
YBachmann: Expanded description.

YBachmann:
> It looks like you've made some changes to address my comments. When you have a chance, could you reply to each review comment to note that you've addressed it, and let me know when this PR is ready for another look?

Hey @jonbinney , 
I replied to each comment and marked it as resolved, please re-open a comment if you think it still needs something changed.
Edit: I Just noticed, that an example yaml and Python launchfile are still missing. I will add those on monday.

jonbinney:
Thanks! One more small request: could you add a comment in the code about why the window size needs to be incremented if it is even? Your explanation was good, just copy it into the code there: "If the window_size is even, the current range, which will be modified cannot be centered in the window." Once you do that and add the example files, this should be ready for me to merge.

YBachmann:
Hello @jonbinney , 
sorry for the late response, I had some urgent things to do this week. 
I added an example `xml` and `python` launchfile and a parameter `yaml` file that's used by the launchfiles. I also added a comment regarding the odd `window_size_` in the code. 
Should this be the last update, I would like to thank you for taking the time to reviewing this PR!

jonbinney (approved):

jonbinney:
Merging - thanks for contributing this @YBachmann !

PR #202, opened 2024-08-21
Using NaN instead of range_max+1 to remove scans in angular_bounds_filter_in_place

bely2803:
## Problem statement
The current implementation of `angular_bounds_filter_in_place` uses `range_max + 1.0` to filter laser scans, which can lead to unintended behaviours when these scans are passed to packages like `AMCL` and `slam_toolbox`.

## Proposed Change
This PR modifies the code to use `NaN` instead of `range_max + 1` for filtering out-of-range scans. Using `NaN` is a more robust approach because it eliminates the need for other packages to verify whether each scan falls within the range specified in the message header.

This change aligns with the existing methodology used in `box_filter`, enhancing consistency across the codebase. We've ported this approach to `angular_bounds_filter_in_place` to ensure more reliable and predictable performance.


jonbinney:
This is a good change! We'll have to be careful how we apply it though. There may be people with robots in the field who depend on the current behavior. Normally we only make user-facing changes along with new ROS distros. I propose the following:

- add a param to this filter called `replace_with_nan` and have it default to false
- when the param is set to true, use NaN, else use range_max + 1. this will preserve the old behavior as default for the moment, but will allow you to start using the (more correct) value
- next time there is a new ROS distro, we'll swap the default value to false. This will still probably break some peoples' robots, but they'll be looking for bugs since it is a new distro.

Additionally, put a comment in the code where the parameter is declared describing this and linking to this PR.

jonbinney (approved):
Looks good!

PR #204, opened 2024-08-27
Dynamic setting of params in speckle filter

jonbinney:
Done previously by berend-kupers, but removed during code review because we thought it needed a separate node inside the filter. Here we use the params_interface_ member variable of the filter base class instead of creating a new node.

The callback needs to be post_set_params and not on_set_params because the changes haven't been finalized yet when on_set_params is called.


jonbinney:
Tests fail on humble because the parameter callback interface was different than - there was no separate "post" set params then. We'll need to either branch  for humble or use preprocessor conditionals to decide which interface to compile for. I'll take another look at this later in the week.

jonbinney:
Builds now, but i want to add a test of the parameter setting before I trust it.

PR #212, opened 2025-03-24
Port ros1 functionalities

jeaninevbrr:
Ports already existing functionality from ros1: 
* inverting range parameter
* parameter to override intensities to either 0 or 1 
* parameter to override range to Nan 
* add reconfigure callback

Related ros1 [commit](https://github.com/ros-perception/laser_filters/commit/3b3c5ac5f5f1d07a61b4ae537a041891bc38e615)
One thing I worry about: Because of [this](https://github.com/ros/filters/blob/658b202d144fb607d82e80356db1897623107c68/include/filters/filter_base.hpp#L171) implementation of getParam() the default  for "filter_override_range" is false. This changes the current default behavior of the filter when no parameters are set. Do you have any suggestions for a workaround?

### DIFF
--- a/include/laser_filters/intensity_filter.h
+++ b/include/laser_filters/intensity_filter.h
@@ -61,18 +61,12 @@ public:
 
   bool configure()
   {
-    lower_threshold_ = 8000.0;
-    upper_threshold_ = 100000.0;
-    disp_hist_ = 1;
-    invert_ = false;
-    filter_override_range_ = true;
-    filter_override_intensity_ = false;
-    getParam("lower_threshold", lower_threshold_);
-    getParam("upper_threshold", upper_threshold_) ;
-    getParam("disp_histogram",  disp_hist_) ;
-    getParam("invert", invert_);
-    getParam("filter_override_range", filter_override_range_);
-    getParam("filter_override_intensity", filter_override_intensity_);
+    getParam("lower_threshold", lower_threshold_, 8000.0);
+    getParam("upper_threshold", upper_threshold_, 100000.0) ;
+    getParam("disp_histogram",  disp_hist_, 1) ;
+    getParam("invert", invert_, false);
+    getParam("filter_override_range", filter_override_range_, true);
+    getParam("filter_override_intensity", filter_override_intensity_, false);
 
     disp_hist_enabled_ = (disp_hist_ == 0) ? false : true;
 
--- a/include/laser_filters/intensity_filter.h
+++ b/include/laser_filters/intensity_filter.h
@@ -55,15 +55,24 @@ public:
   double upper_threshold_ ;
   int disp_hist_ ;
   bool disp_hist_enabled_;
+  bool invert_;
+  bool filter_override_range_;
+  bool filter_override_intensity_;
 
   bool configure()
   {
     lower_threshold_ = 8000.0;
     upper_threshold_ = 100000.0;
     disp_hist_ = 1;
+    invert_ = false;
+    filter_override_range_ = true;
+    filter_override_intensity_ = false;
     getParam("lower_threshold", lower_threshold_);
     getParam("upper_threshold", upper_threshold_) ;
     getParam("disp_histogram",  disp_hist_) ;
+    getParam("invert", invert_);
+    getParam("filter_override_range", filter_override_range_);
+    getParam("filter_override_intensity", filter_override_intensity_);
 
     disp_hist_enabled_ = (disp_hist_ == 0) ? false : true;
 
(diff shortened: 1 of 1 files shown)


jonbinney:
If you do:
```
bool foo = true; // this is your default
this->getParam("foo", foo)
```
then if foo wasn't set on the param server, then it should still be true, right?

jeaninevbrr:
Yes, that's what I thought would happen, but my value bacomes false after calling getParam().  I think this happens in 
[here](https://github.com/ros/filters/blob/658b202d144fb607d82e80356db1897623107c68/include/filters/filter_base.hpp#L124).
Because the parameter does not exist in the ros parameter server it ends up  in the !has_parameter context, and thus setting the default.
The getParam(name, value) uses the value input only to store the retrieved parameter, not to set a default

jonbinney:
Great catch  - I vaguely remember that the parameter reading code was a bit wacky because it was trying to emulate ROS1 behavior to make porting filters easy, but ROS2 requires explicit parameter declaration (at least for read only params) and that made things tricky. I'll take a closer look later but one solution may be to implement the TODO here: https://github.com/ros/filters/blob/658b202d144fb607d82e80356db1897623107c68/include/filters/filter_base.hpp#L144 so that callers know the parameter wasn't set before and can use their default. Does that sound right to you?

jeaninevbrr:
I'm not sure if the TODO is the right place to fix it. You could check there weather the value_out already has a value. But that seems a bit wack. Why don't we add the default_value as input for getParam with its own default?:
```
  bool getParam(const std::string & name, bool & value, bool default_value = false)
  {
    return getParamImpl(name, rcl_interfaces::msg::ParameterType::PARAMETER_BOOL, default_value, value);
  }
```
This, way we don't break the current default behavior, but we can give defaults in the filter class:
```
    getParam("lower_threshold", lower_threshold_, 8000.0) ;
    getParam("upper_threshold", upper_threshold_, 100000.0) ;
    getParam("disp_histogram",  disp_hist_, 1) ;
    getParam("invert", invert_, false);
    getParam("filter_override_range", filter_override_range_, true);
    getParam("filter_override_intensity", filter_override_intensity_, false);
```

jonbinney:
I think you're right about not changing existing functionality but i like the idea of passing in the default. Want to do me a favor and PR that change to filters?

jeaninevbrr:
yeah sure!

jonbinney:
Wow that was quick! Thank you!

jonbinney:
CI is failing for some reason. From the logs:
```
1: /__w/laser_filters/laser_filters/src/laser_filters/test/test_scan_filter_chain.cpp:69: Failure
1: The difference between a[i] and b[i] is nan, which exceeds 1e-6, where
1: a[i] evaluates to 0.10000000149011612,
1: b[i] evaluates to nan, and
1: 1e-6 evaluates to 9.9999999999999995e-07.
1: /__w/laser_filters/laser_filters/src/laser_filters/test/test_scan_filter_chain.cpp:[69](https://github.com/ros-perception/laser_filters/actions/runs/14039141787/job/39305033064?pr=212#step:6:70): Failure
1: The difference between a[i] and b[i] is nan, which exceeds 1e-6, where
1: a[i] evaluates to 9,
1: b[i] evaluates to nan, and
1: 1e-6 evaluates to 9.9999999999999995e-07.
```

jonbinney:
Any idea why there's a `nan` where there should be?

jeaninevbrr:
Yes, it will be solved by setting the defaults correctly. 
Because filter_override_range_ is false by default, no ranges will be overwritten with a nan. Output = input when both filter_override_range_ and filter_override_intensity are false. so the test fails.  The PR on filters fixes this, but i guess then these tests should wait on a new filters release?

jonbinney:
Ah, yes, we should get the filters PR merged and release first. I'll try to be quick about it though.

jonbinney:
Issue for the filters release: https://github.com/ros/filters/issues/86

jeaninevbrr:
Can you merge? :)

jonbinney:
I think this is ready for merging, but I want to take one more look through it today or tomorrow since it's been a while and I've forgotten the details.

jonbinney (approved):
Looks good. Most of the new parameters default to false and keep the old behavior, and overriding the ranges when filtering seems to make sense.

jonbinney:
Thanks @jeaninevbrr! Do you have more PRs to send? Did I miss any that you already sent?

jeaninevbrr:
Nope all done! Now we wait for the new release :). Thanks for the pleasant cooperation!

jonbinney:
And thank you for all the great contributrions! I just released version `2.0.9` to humble and jazzy with all your new stuff. It'll take some time from them to work their way through the buildfarm.

- humble: https://github.com/ros/rosdistro/pull/46765
- jazzy: https://github.com/ros/rosdistro/pull/46766

Kilted and rolling are now released from "kilted" and "rolling" branches because of some non backwards compatible changes in other packages, so i'll have to port your changes across to those before I release on those distros. I should get to it within the next week.

PR #213, opened 2025-03-24
Added params_prefix to reconfigure callback

jeaninevbrr:
To be able to use the reconfigure callback properly the "params_prefix_" is necessary to compare correctly. Otherwise it would throw "unknown parameter" when I try to update filter#.params.parameter_a because it looks for parameter_a

### DIFF
--- a/include/laser_filters/polygon_filter.h
+++ b/include/laser_filters/polygon_filter.h
@@ -309,17 +309,19 @@ protected:
 
     for (auto parameter : parameters)
     {
-      if(parameter.get_name() == "polygon"&& parameter.get_type() == rclcpp::ParameterType::PARAMETER_STRING){
+      if(logging_interface_ != nullptr)
+          RCLCPP_INFO_STREAM(logging_interface_->get_logger(), "Update parameter " << parameter.get_name().c_str()<< " to "<<parameter);
+      if(parameter.get_name() == param_prefix_+"polygon"&& parameter.get_type() == rclcpp::ParameterType::PARAMETER_STRING){
         std::string polygon_string = parameter.as_string();
         polygon_ = makePolygonFromString(polygon_string, polygon_);
       }
-      else if(parameter.get_name() == "polygon_frame" && parameter.get_type() == rclcpp::ParameterType::PARAMETER_STRING){
+      else if(parameter.get_name() == param_prefix_+"polygon_frame" && parameter.get_type() == rclcpp::ParameterType::PARAMETER_STRING){
         polygon_frame_ = parameter.as_string();
       }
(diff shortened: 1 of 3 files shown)


jonbinney:
This seems correct but I'll need to sit down sometime this week and debug. @jeaninevbrr if you've got more PRs for filters or laser_filters, could you hold off until the first 3 go through the review process? These are great and I really appreciate your work to improve these packages but I'm feeling overwhelmed :-)

jeaninevbrr:
Yes, I have some more coming indeed ;-), including some bug fixes. We are a bit behind merging stuff back to you. But i'll keep the rest in branches for now.

include/laser_filters/polygon_filter.h
```diff
@@ -309,17 +309,19 @@ class LaserScanPolygonFilterBase : public filters::FilterBase<sensor_msgs::msg::
 
     for (auto parameter : parameters)
     {
-      if(parameter.get_name() == "polygon"&& parameter.get_type() == rclcpp::ParameterType::PARAMETER_STRING){
+      if(logging_interface_ != nullptr)
+          RCLCPP_INFO_STREAM(logging_interface_->get_logger(), "Update parameter " << parameter.get_name().c_str()<< " to "<<parameter);
```
jonbinney: Did you mean to leave this in? Probably best to take it out, or at least change it to DEBUG

jeaninevbrr:
It was already in the speckle_filter, so for consistentcy i added it to the other filters as well. But I am fine with making it debug.

include/laser_filters/polygon_filter.h
jeaninevbrr: It was already in the speckle_filter, so for consistentcy i added it to the other filters as well. But I am fine with making it debug.

jonbinney:
This PR is fixing a clear bug.... but I'm a bit worried that some people may be relying on the incorrect behavior, and this could break their robots. I'm trying to decide whether we need to roll this out slowly (maybe put the behavior behind an option that defaults to false and outputs a warning for now, and then swap the param to default true in rolling and new distros)

Thoughts?

EDIT: "new releases" -> "new distros"

jeaninevbrr:
hmm, i get your point. Since it is a clear bug, I would hope that people relying on it would open a pull request :-P. 
In my opinion it is fine to merge. It is only the parameter reconfigure, not the functionality of the filter itself. I can not really estimate though how many users are changing there filter parameters during runtime, other then for debugging purposes. And it gives a "Unknown parameter" warning when this goes wrong. We could extend that warn message maybe?

jonbinney:
Yeah, after sleeping on it I agree it is reasonable to merge.

PR #215, opened 2025-04-02
Use correct footprint subscriber callback for static polygon filter

jeaninevbrr:
Footprint subscriber for static polygon filter was only using the callback function of the base class and not the overrided function in the child class.

### DIFF
--- a/include/laser_filters/polygon_filter.h
+++ b/include/laser_filters/polygon_filter.h
@@ -234,7 +234,7 @@ public:
               std::bind(&LaserScanPolygonFilterBase::reconfigureCB, this, std::placeholders::_1));
 
     std::string polygon_string;
-    if(!filters::FilterBase<sensor_msgs::msg::LaserScan>::getParam(std::string("footprint_topic"), footprint_topic_, "base_footprint_exclude"))
+    if(!filters::FilterBase<sensor_msgs::msg::LaserScan>::getParam(std::string("footprint_topic"), footprint_topic_, false, "base_footprint_exclude"))
     {
       RCLCPP_WARN(logging_interface_->get_logger(), "Footprint topic not set, assuming default: base_footprint_exclude");
     }
--- a/include/laser_filters/polygon_filter.h
+++ b/include/laser_filters/polygon_filter.h
@@ -236,15 +236,14 @@ public:
     std::string polygon_string;
     invert_filter_ = false;
     polygon_padding_ = 0;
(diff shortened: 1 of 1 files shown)


jeaninevbrr:
I will fix the tests first :)

include/laser_filters/polygon_filter.h
```diff
@@ -236,15 +236,14 @@ class LaserScanPolygonFilterBase : public filters::FilterBase<sensor_msgs::msg::
     std::string polygon_string;
     invert_filter_ = false;
     polygon_padding_ = 0;
-    std::string footprint_topic;
-    if(!filters::FilterBase<sensor_msgs::msg::LaserScan>::getParam(std::string("footprint_topic"), footprint_topic))
+    if(!filters::FilterBase<sensor_msgs::msg::LaserScan>::getParam(std::string("footprint_topic"), footprint_topic_))
```
jonbinney: Could you update this to use the snazzy new default parameter for `FilterBase::getParam`? Then you can skip the Check on line 244 below.

jonbinney (changes requested):
Looks good except for one small change.

include/laser_filters/polygon_filter.h
jeaninevbrr: Yes, done. Also deleted invert_filter_ and polygon_padding_ because these are already set by getparam's default.

jeaninevbrr: But, I am looking at this now and realizing that getParam never returns false. So this whole flow of checking parameters does not make sense right?

jonbinney:
I'm having trouble getting CI to re-run since it has been more than 30 days since the last run. @jeaninevbrr could you merge the ros2 branch into this branch and push again top your fork? I think that should do the trick. Same thing for https://github.com/ros-perception/laser_filters/pull/219

jeaninevbrr:
I guess we have to wait for rolling and jazzy sync as well

jonbinney:
Yeah - but we're close!

include/laser_filters/polygon_filter.h
jonbinney: Yeah....

jonbinney (approved):
Looks good, and all tests passing now!

jonbinney:
There's new branches for rolling and kilted now, I'll backport (forwardport?) this change to those.

jonbinney:
@Mergifyio copy kilted

PR #216, opened 2025-04-02
Added window size check to prevent segfault in speckle filter

jeaninevbrr:
Was already present in [ros1 ](https://github.com/ros-perception/laser_filters/blob/noetic-devel/src/speckle_filter.cpp#L83).


jonbinney (approved):
LGTM!

PR #217, opened 2025-04-02
Added heartbeat diagnostics

jeaninevbrr:
Since the segfault crash fixed with #216, we added heartbeat diagnostics to the nodes, so the robot can go into error state when a node has crashed. What do you think about it?
 
 PS: No pressure on solving all these PR's quickly, lets just keep a steady pace :)

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -14,6 +14,8 @@
   <build_depend>ament_cmake_auto</build_depend>
 
   <depend>angles</depend>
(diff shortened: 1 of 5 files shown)


jonbinney:
Greatl - diagnostic_msgs are one of my favorite parts of ROS. I'll review the code this weekend.

jonbinney (approved):
Looks good, this is a nice improvement!

PR #219, opened 2025-04-17
make reconfigurable parameters writable

jeaninevbrr:
As a result of https://github.com/ros/filters/pull/83, parameters need to be set to read_only = false in order to use the reconfigure callback

### DIFF
--- a/include/laser_filters/polygon_filter.h
+++ b/include/laser_filters/polygon_filter.h
@@ -237,7 +237,7 @@ public:
     invert_filter_ = false;
     polygon_padding_ = 0;
     std::string footprint_topic;
-    if(!filters::FilterBase<sensor_msgs::msg::LaserScan>::getParam(std::string("footprint_topic"), footprint_topic))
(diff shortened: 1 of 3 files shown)


jonbinney:
CI is still getting version `2.1.0` of filters from apt, which is why it is still failing to compile. Looks like jazzy and humble had apt package syncs last on april 24/25, so i'd expect the next one to be in a week or so.

jonbinney:
Sync is happening in a couple days, then this should pass CI

jonbinney (approved):
LGTM

jonbinney:
@jeaninevbrr could you resolve the merge conflicts?

jeaninevbrr:
these rolling test errors I cannot really place..

jonbinney:
Ah, those are the result of some breaking changes made upstream in the message_filters package. We're now releasing rolling from the new "rolling" branch which has fixes to deal with that, so I just removed rolling from the CI for the ros2 branch. Could you merge main into this PR branch to get the update CI workflow?

jeaninevbrr:
Can you merge? :)

jonbinney:
I can! :-)

PR #227, opened 2025-07-01
Fix compile warning in speckle filter

jonbinney:


jonbinney:
@mergifyio backport rolling

PR #228, opened 2025-07-18
Port PR 215 to rolling

jonbinney:


jonbinney:
@Mergifyio copy kilted

PR #232, opened 2025-08-19
Feat: angular_bounds_filter_in_place dynamic parameters and new enabled parameter

bely2803:
## Problem statement
In our application, we need to toggle angular bounds filters at runtime to handle physical configuration changes in our mobile platform. The current implementation does not provide this feature.

## Proposed changes
A new enabled parameter is added, when toggled to false the scan will simply copy the input_scan into the output_scan.
The reconfigureCB() method is overridden to handle dynamic reconfiguration of the filter's parameters.

### DIFF
--- a/include/laser_filters/angular_bounds_filter_in_place.h
+++ b/include/laser_filters/angular_bounds_filter_in_place.h
@@ -48,13 +48,14 @@ namespace laser_filters
       double lower_angle_;
       double upper_angle_;
       bool replace_with_nan_;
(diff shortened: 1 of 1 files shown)


jonbinney:
Instead of making this filter more complicated by adding an on/off toggle, could you use a [mux](https://github.com/ros-tooling/topic_tools?tab=readme-ov-file#mux) node to toggle between the scan and the filtered scan when the configuration changes?

PR #234, opened 2025-09-01
Set hardware id

### LINKED ISSUE #233: No HW_ID was set
The diagnostic has been introduced, but since the hardware ID has not been set, the following warning appears at startup.

```bash
[WARN][laser.speckle_filter]: diagnostic_updater: No HW_ID was set. This is probably a bug. Please report it. For devices that do not have a HW_ID, set this value to 'none'. This warning only occurs once all diagnostics are OK. It is okay to wait until the device is open before calling setHardwareID.
```

Tacha-S:
I specified the hardware ID so that the warning no longer appears.

- fix #233

### DIFF
--- a/src/scan_to_scan_filter_chain.cpp
+++ b/src/scan_to_scan_filter_chain.cpp
@@ -44,7 +44,8 @@ ScanToScanFilterChain::ScanToScanFilterChain(
 {
   // Heartbeat diagnostics
   diagnostic_updater_.add(heartbeat_diagnostics_);
-  
+  diagnostic_updater_.setHardwareID("laser_filters");
+
   // Configure filter chain
   filter_chain_.configure(
     "",


jonbinney:
@Tacha-S  Thanks for this!  It seems like there are three reasonable options for hardware_id in a node that doesn't directly correspond to a hardware device (since the laser driver is in a separate node and package):
1. "none"
2. "<package_name>"  (which you use here)
3. "<node_name>" (more unique than package_name, since there may be multiple laser_filter nodes)
What are your thoughts on which option is best?

Tacha-S:
I think the recommendation from diagnostic_updater is indeed option 1, `"none"`.
If we follow that strictly, option 1 would be best.

However, looking at other packages (such as navigation2), it seems they often use a package-specific identifier, so I chose the package name to make it clearer.
If clarity is the priority, I think option 2 would be the best choice.

jonbinney (approved):
LGTM

jonbinney:
Makes sense. If using the package name causes confusion for people running multiple laser_filters nodes, we can think about changing it later.

Thanks again!

jonbinney:
@Mergifyio backport rolling kilted

PR #239, opened 2025-09-04
Backport #212

jonbinney:
-Make intensity filter reconfigurable
-Allow overriding of intensity and range values when filtering


jonbinney:
@mergifyio backport kilted

PR #242, opened 2025-09-04
Fix compile warning in speckle filter (Backport #227 to rolling)

jonbinney:


jonbinney:
@mergifyio backport kilted

PR #255, opened 2026-03-31
Enable qos overriding for the output topic

Hugal31:
Enable QoS overriding for the publisher.

This allows to configure the publisher qos via the `qos_overrides./scan_filtered.publisher.*` parameters.
This makes the `scan_filtered_history_depth` parameter a bit redundant.

Reason: I am using rmw_zenoh, and in Zenoh the publisher decides of the final reliability. I want my scans to be in best_effort.


jonbinney (approved):
Looks good!

jonbinney:
Thanks for this @Hugal31 !

PR #257, opened 2026-04-15
Configure the tf buffer timer creator in scan_to_scan_filter_chain

### LINKED ISSUE #256: Error "timer interface not set before call to waitForTransform" when setting tf_message_filter_target_frame
# Description

When using `tf_message_filter_target_frame`, the scan_to_scan_filter_chain crashes with the following error:
```
terminate called after throwing an instance of 'tf2_ros::CreateTimerInterfaceException'
  what():  timer interface not set before call to waitForTransform
```

This is because tf2_ros::MessageFilter uses tf2_ros::Buffer `waitForTransform` with a callback. For that, it needs to have been configured by calling setCreateTimerInterface first.

# Reproduce

Run in parallel:
```
> ros2 run laser_filters scan_to_scan_filter_chain --ros-args -p tf_message_filter_target_frame:=foobar
> ros2 topic pub /scan sensor_msgs/msg/LaserScan 'header: {frame_id: "nop"}'
```

Hugal31:
Configure the tf buffer by calling setCreateTimerInterface so the message filter can call waitForTransform.

How should I go about making a backport for jazzy?

Closes #256

### DIFF
--- a/src/scan_to_scan_filter_chain.cpp
+++ b/src/scan_to_scan_filter_chain.cpp
@@ -30,6 +30,8 @@
 
 #include "scan_to_scan_filter_chain.hpp"
 
+#include <tf2_ros/create_timer_ros.hpp>
+
 
 // Constructor
 ScanToScanFilterChain::ScanToScanFilterChain(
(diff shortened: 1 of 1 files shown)


jonbinney:
Does this approach work for you? I vaguely remember a problem with new transforms not being added to the buffer if the filter is blocking on waitfortransform, because that blocks the executor callback that would handle adding tf messages to the buffer.

Hugal31:
Yes does work, I sometimes have to wait a bit on lookupTransform in my filters because of #258.

The TransformListener is spinning its own thread and executor, ~~and even without it I don't see how that would block the executor.~~

Edit: indeed if the TransformListener didn't spin a dedicated thread if a filter tries to wait for a new one, that would hang everything. However if the tf buffer is shared between the filter chain and each filter, the tf2_ros::MessageFilter would have removed the need to wait for the transform, but you never know when a filter would want to wait for another frame for some reason.

jonbinney:
Thanks @Hugal31  - I'll dig into this and the other issues you brought up. I'm busy this week, but I should be able to get to them mid next week.

wjwwood:
btw, this needs a change to avoid a deprecated (and now removed on rolling) API, I opened a pr: https://github.com/Hugal31/laser_filters/pull/1

PR #259, opened 2026-04-16
Re-use the main filter node for the tf2_ros::MessageFilter TransformListener

Hugal31:
The TransformListener [recommends to re-use a node](https://github.com/ros2/geometry2/blob/51cc6451e49385ff03a324d23641910972764414/tf2_ros/include/tf2_ros/transform_listener.hpp#L103-L105) when possible.

This avoids spawning a new node uselessly.

By the way, I think we could also pass spin_thread=false in this specific use case, since we don't use the timeout mechanism.

### DIFF
--- a/src/scan_to_cloud_filter_chain.cpp
+++ b/src/scan_to_cloud_filter_chain.cpp
@@ -45,7 +45,7 @@ ScanToCloudFilterChain::ScanToCloudFilterChain(
   diagnostic_updater_(this),
   laser_max_range_(DBL_MAX),
   buffer_(this->get_clock()),
-  tf_(buffer_),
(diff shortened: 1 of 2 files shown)


jonbinney:
Interesting.... in the old code, after launching, do you see the extra node running (e.g. with `ros2 node list`)?

Hugal31:
Yes

```
> ros2 run laser_filters scan_to_scan_filter_chain --ros-args -p tf_message_filter_target_frame:=foobar &
> ros2 node list
/scan_to_cloud_filter_chain
/transform_listener_impl_63ec3974c7c0
```

jonbinney (approved):
Looks good!

PR #263, opened 2026-07-20
[ROS-O] drop obsolete boost system component

v4hn:
The component has been empty for a very long time and was finally dropped from cmake.
Fails in 26.04. without this cleanup.


jonbinney (approved):
LGTM

PR #248, opened 2025-09-25
Create unique nodes for some filters

jonbinney:
This is an attempt to fix #231

Issue #231, opened 2025-08-19
topic subscription in polygon filter is not working


YannickdeHoop:
The topic subscription to get a polygon is not working in the polygon filter. I think this is happening because the polygon filter is inherited from rclcpp::Node, even though the filter chain itself is also a node. This results in undefined behaviour. If you start a polygon filter in a filter chain, it will complain about two nodes with the same name

jonbinney:
I think it currently creates its own node just to publish the polygon: https://github.com/ros-perception/laser_filters/blob/5fc32268c5e18ec5f867099aadb5e752601bfb38/include/laser_filters/polygon_filter.h#L259

I've been trying to avoid passing the entire node handle into each filter in FilterBase in the filters package, because that would allow them to do things that could interfere with each other. Currently filters can create parameters and respond to parameter updates via an interface provided by FilterBase, and also log via the inherited `logging_interface_` member variable of FilterBase. But they can't publish or subscribe to topics without creating their own node since they don't have a node handle.

This issue of needing multiple nodes, which causes problems, seems like a point in favor of giving up on isolating the filters from each other, and just passing a node handle to each of them. See this discussion on the filters package for a discussion of this option: https://github.com/ros/filters/issues/91

It'd be great to have another person contributing their perspective to that issue. I'm also open to other ideas about how to design a filters API in ros2 that allows for most common use cases while still keeping filters easy to write and maintain, and not providing too many footguns.

jonbinney:
Also there may be a more specific solution that could fix this particular problem with the polygon filter. In ros2 you can have multiple nodes in one executable, so I don't see why it would cause a problem. Maybe the multiple node thing is a red herring and this is related to namespaces?

YannickdeHoop:
i dont mean the publisher but subscriber at line 364. This is a subscriber of the footprint which can be used instead of a static footprint in the parameters. 
Did you ever take a look at ROS 2 components? [components_ROS2](https://docs.ros.org/en/rolling/Tutorials/Intermediate/Composition.html)
Components are separate nodes which can be launched dynamically. They can have intra process communication so you dont have overhead. In this way you can create kind of a pipeline, like nodelets in ros1. 

noise_filter
```
#include <rclcpp/rclcpp.hpp>
#include <sensor_msgs/msg/laser_scan.hpp>

class NoiseFilter : public rclcpp::Node
{
public:
  NoiseFilter(const rclcpp::NodeOptions & options)
  : Node("noise_filter", options)
  {
[...]

PR #266, opened 2026-08-14
Give box/footprint/polygon filters unique private node names

### LINKED ISSUE #214: Creating multiple nodes when changing node name
When I set name in launch file, I get the following error

```bash
[scan_to_scan_filter_chain-5] [WARN] [1743186960.346687424] [rcl.logging_rosout]: Publisher already registered for node name: 'laser_filter'. If this is due to multiple nodes with the same name then all logs for the logger named 'laser_filter' will go out over the existing publisher. As soon as any node with that name is destructed it will unregister the publisher, preventing any further logs for that name from being published on the rosout topic.
```

```python
    laser_filter_node = Node(
        package="laser_filters",
        executable="scan_to_scan_filter_chain",
        name="laser_filter", # Removing this fix the problem
        parameters=[laser_filter_config],
    )
```

rafal-gorecki:
Fixes #214.

`box_filter`, `footprint_filter` and `polygon_filter` each spin up their own
internal node (for TF, and for polygon_filter also pub/sub/param callbacks).
That internal node used a fixed name, so a process-wide bare
`-r __node:=<name>` remap (e.g. `Node(name="laser_filter")` in a launch file)
silently renames it too, colliding with the parent filter-chain node and
producing `Publisher already registered for node name: 'laser_filter'`,
exactly as reported here and in https://github.com/ros-perception/laser_filters/issues/214#issuecomment-5291590730.

Each internal node now gets a name unique to the filter instance, pinned via
its own local `-r __node:=<name>` argument — that survives a process-wide
bare remap since node-specific rules take precedence over the wildcard one.

This follows the same approach @jonbinney used in #248 for box_filter/
polygon_filter (opened against #231), and extends it to footprint_filter,
which #248 doesn't cover but is affected by the same root cause.

Tested against the exact repro from the issue (scan_to_scan_filter_chain,
`--remap __node:=laser_filter`, box_filter via params file): the
"Publisher already registered" warning and the node-name collision are
gone, and the box filter's internal node keeps a unique name in
`ros2 node list`.

### DIFF
--- a/include/laser_filters/box_filter.h
+++ b/include/laser_filters/box_filter.h
@@ -73,7 +73,7 @@ class LaserScanBoxFilter : public filters::FilterBase<sensor_msgs::msg::LaserSca
     {
       node_ = getUniqueNode("box_filter", this);
       buffer_ = std::make_shared<tf2_ros::Buffer>(node_->get_clock());
-      tf_ = std::make_shared<tf2_ros::TransformListener>(*buffer_);
+      tf_ = std::make_shared<tf2_ros::TransformListener>(*buffer_, node_);
 
       up_and_running_ = true;
       double min_x, min_y, min_z, max_x, max_y, max_z;
--- a/include/laser_filters/box_filter.h
+++ b/include/laser_filters/box_filter.h
@@ -56,20 +56,25 @@
 typedef tf2::Vector3 Point;
 
 #include <laser_geometry/laser_geometry.hpp>
-#include <rclcpp_lifecycle/lifecycle_node.hpp>
+
+#include "laser_filters/util.h"
 
 namespace laser_filters
 {
 /**
  * @brief This is a filter that removes points in a laser scan inside of a cartesian box.
  */
-class LaserScanBoxFilter : public filters::FilterBase<sensor_msgs::msg::LaserScan>, public rclcpp_lifecycle::LifecycleNode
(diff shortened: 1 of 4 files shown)


rafal-gorecki:
A note on why this also drops `rclcpp_lifecycle::LifecycleNode` in favor of a plain `rclcpp::Node` for the private per-filter node (box/footprint/polygon):

These classes never actually engage the lifecycle state machine — no `on_configure`/`on_activate` overrides, and nothing anywhere in the codebase calls `trigger_transition()` on them. `configure()` here is just `filters::FilterBase`'s own interface method, unrelated to the lifecycle callback of the same name.

Despite that, the resulting node still advertises itself as a fully manageable lifecycle node (`~/change_state`, `~/get_state`, `~/transition_event` — confirmed via `ros2 lifecycle nodes`). Since nothing ever spins its executor, any real interaction with those services just hangs forever (confirmed: `ros2 lifecycle get <node>` never returns). So today it's a node that *looks* lifecycle-managed to any tooling that discovers it that way, but isn't — pure surface area for zero benefit. A plain `rclcpp::Node` gives these filters exactly what they use (logger, clock, pub/sub, param callback) without that trap, and matches the approach @jonbinney already took in #248 for box_filter/polygon_filter.

rafal-gorecki:
@jonbinney please review

